PR #18, opened 2020-09-26
Fix crates.io Badges

zalo:
`rapier` is published with `rapier2d` and `rapier3d` variants, so this change adds badges for both and links to the correct pages.


sebcrozet:
Thanks!

PR #33, opened 2020-10-10
Fix stack overflow caused by the WQuadtree

sebcrozet:
The WQuadtree construction function would overflow the stack whenever more than 4 AABBs with the same center are provided. Moreover, this same configuration would cause a subtraction underflow.

This PR fixes both bugs.

### DIFF
--- a/examples3d/debug_boxes3.rs
+++ b/examples3d/debug_boxes3.rs
@@ -17,22 +17,26 @@ pub fn init_world(testbed: &mut Testbed) {
     let ground_size = 100.1;
     let ground_height = 0.1;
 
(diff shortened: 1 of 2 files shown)


robert-hrusecky (approved):
Looks good to me!

Bobox214:
I confirm that this fixes panics I add in the form of :
```rust
thread 'Compute Task Pool (3)' panicked at 'attempt to subtract with overflow', /home/boris/.cargo/registry/src/github.com-1ecc6299db9ec823/rapier2d-0.2.0/src/geometry/wquadtree.rs:542:21
```

PR #34, opened 2020-10-11
Add wake_up parameter to rigidbody methods

### LINKED ISSUE #5: Rigid bodies require manually waking when applying forces/impulses/setting velocity directly
I am having to `rb.wake_up()` whenever I apply an impulse or change velocity directly. This doesn't seem like the desired behavior

robert-hrusecky:
Add boolean parameter to rigid body methods as discussed in #5 
Closes #5

### DIFF
--- a/src/dynamics/rigid_body.rs
+++ b/src/dynamics/rigid_body.rs
@@ -272,65 +272,102 @@ impl RigidBody {
      * Application of forces/impulses.
      */
     /// Applies a force at the center-of-mass of this rigid-body.
-    pub fn apply_force(&mut self, force: Vector<f32>) {
+    pub fn apply_force(&mut self, force: Vector<f32>, wake_up: bool) {
         if self.body_status == BodyStatus::Dynamic {
             self.linacc += force * self.mass_properties.inv_mass;
+        }
+
+        if wake_up {
+            self.wake_up(false);
         }
     }
 
     /// Applies an impulse at the center-of-mass of this rigid-body.
-    pub fn apply_impulse(&mut self, impulse: Vector<f32>) {
+    pub fn apply_impulse(&mut self, impulse: Vector<f32>, wake_up: bool) {
         if self.body_status == BodyStatus::Dynamic {
             self.linvel += impulse * self.mass_properties.inv_mass;
         }
+
+        if wake_up {
+            self.wake_up(false);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


src/dynamics/rigid_body.rs
```diff
@@ -272,65 +272,102 @@ impl RigidBody {
...
+    ) {
         let torque_impulse = (point - self.world_com).gcross(impulse);
-        self.apply_impulse(impulse);
-        self.apply_torque_impulse(torque_impulse);
+        self.apply_impulse(impulse, false);
+        self.apply_torque_impulse(torque_impulse, false);
+
+        if wake_up {
+            self.wake_up(false);
+        }
```
sebcrozet, line 370:
```suggestion
        self.apply_impulse(impulse, wake_up);
        self.apply_torque_impulse(torque_impulse, wake_up);
```

Same remark as before.

src/dynamics/rigid_body.rs
```diff
@@ -272,65 +272,102 @@ impl RigidBody {
...
+    pub fn apply_force_at_point(&mut self, force: Vector<f32>, point: Point<f32>, wake_up: bool) {
         let torque = (point - self.world_com).gcross(force);
-        self.apply_force(force);
-        self.apply_torque(torque);
+        self.apply_force(force, false);
+        self.apply_torque(torque, false);
+
+        if wake_up {
+            self.wake_up(false);
+        }
```
sebcrozet, line 354:
```suggestion
        self.apply_force(force, wake_up);
        self.apply_torque(torque, wake_up);
```

I understand the approach you took was to avoid making two wake up calls. But this should not have any noticeable performance impact, so passing the `wake_up` parameter will keep things easier to maintain.

sebcrozet (changes requested):
Thank you for this PR!
- All the `self.wake_up(false)` should be `self.wake_up(true)`. This will perform a "strong" wake-up, casing the rigid-body to stay awake for a couple of frames.
- All the `if wake_up { ... }` should happen only if `self.body_status == BodyStatus::Dynamic`.
- The `apply_*_at_point()` should just forward the `wake_up` parameter to the methods it calls. This will make maintainance simpler.

sebcrozet (approved):

sebcrozet:
Thanks!

PR #35, opened 2020-10-12
fix panic from usize being set to -1

ashneverdawn:
When center[dim] > split_point[dim] is true for all indices, at the end of the last iteration ilast is set to -1 which causes a panic because it is unsigned. This PR fixes that issue.


ashneverdawn:
Not sure what this failing test is about?

robert-hrusecky:
I believe this panic is already taken care of by #33
CI check is an ongoing issue with PRs from forks

sebcrozet:
@ashneverdawn Thank you for this PR! I will close it because this fix was already part of #33.

Issue #32, opened 2020-10-08
Make shapes more extensible / Add support for user-defined shapes


robert-hrusecky:
Currently it appears that the set of possible shapes that we can deal with are present in a single `Shape` enum. Perhaps the extensibility of this design can be improved. There appears to be some thoughts of this for the future: there is a `ContactDispatcher` trait, but its one implementation, `DefaultContactDispatcher`, is hard-coded into the narrow-phase.

I'm wondering what the goals are for user defined shapes and what the plan is? I would guess since we are not using `ncollide` in full, we want to avoid dynamic dispatch for performance reasons. We could look into options using traits instead. I would like to have this feature for the game I am working on, and I would love to help add it into Rapier. Having this feature would probably also help us make progress faster on issues like #29 and #27.

sebcrozet:
Custom shapes is something I'd like to allow at some point but some big changes need to happen before that:
1. We need to move all the collision-detection stuffs out of Rapier. I've been working on it lately but it will probably not be completed before 2 or 3 months (because this requires some significant modifications of the way ncollide works).
2. I don't think there is any other options than using a `Box<Shape>` trait-object (but I might be wrong). In that case, we need to see what we can do for serializing them: either we `#[serde(skip)]` custom shapes altogether, or we find another solution, based on [erased_serde](https://docs.serde.rs/erased_serde/index.html) perhaps?
3. The `DefaultContactDispatcher` should no longer be hard-coded into the narrow-phase. But that should be dealt with once we complete the modification for 1.

robert-hrusecky:
Hi! Glad to here this is part of the plan!

Another option instead of `Box<dyn Shape>` would be to use type parameters. I'm envisioning usage that would look like this:
```rust
use rapier2d::geometry::collider::DefaultShape;
#[derive(Shape)] // macro that handles all the boilerplate dispatching code
pub enum MyShape {
    Default(DefaultShape),
    MyShapeA(MyShapeA),
    MyShapeB(MyShapeB),
}
```
Here the current `Shape` enum is renamed to `DefaultShape`, and `Shape` is now a trait. `DefaultShape`, `MyShapeA`, and `MyShapeB`, and `MyShape` all implement `Shape`. Rapier types can then be parameterized by `MyShape` in user code. This should make serialization a bit easier. It would also open up the option of excluding some of the default shapes if they are never used in the simulation.

sebcrozet:
Using type parameters is not an option because that would result in:
[...]

Issue #29, opened 2020-10-06
Support rounded corners in cuboid colliders


stephanemagnenat:
Unity colliders support small [rounded corners](https://docs.unity3d.com/ScriptReference/BoxCollider2D-edgeRadius.html), which Box2D/Planck-JS do not support, and that are important when making 2-D games when a character moves over several platforms. Without that, numerical errors might spuriously block the character even if platforms are at the same altitude. If the platforms are static, they can be pre-processed into a single shape, but if platforms are moving, they cannot, yet in some configuration they can touch and the character should be able to move smoothly from one to another. The edge radius of Unity box colliders solves that.

As far as I can see, currently Rapier's cuboid colliders do not support that. It would be nice to support that.

sebcrozet:
Characters are often represented by capsules so they should not suffer from this problem too much.
But yeah, adding rounded corners would be nice. This is something that can be handled for free for arbitrary convex polyhedron (because they already require a generic algorithm that supports rounded corners out-of-the-box). For simple shapes like triangles, and cuboids, we may need some extra work to make it efficient.

sebcrozet:
Rapier will support rounded corners for all convex shapes (including cuboid) in its next version, scheduled to be released at the end of January 2021.

PR #41, opened 2020-10-26
Add cylinder and cone support + use a trait-object for shapes.

sebcrozet:
- Add support for cylinder, cone, and rounded cylinders. Adding rounded cylinders was an opportunity to start experimenting with a way to support #29.
- Shapes are now trait-objects instead of an enum. This is a first work towards making #32.
- Add collision-detection between capsules and triangle meshes.


sebcrozet:
Benchmarks show that the switch from an enum to a trait-object for shapes did not have any performance impact.

PR #44, opened 2020-10-27
Added user-implementable traits for collision/proximity pair filtering.

### LINKED ISSUE #3: Add collision exceptions.
Add the ability to add collision exceptions between the two or more rigid bodies.

### LINKED ISSUE #15: Consider allowing contacts between static/kinematic bodies.
Currently, the detection of collisions between two static bodies, two kinematic bodies, or one static and one kinematic bodies, will be ignored (the contact points won't even be computed).

It may be useful to have a way to tell Rapier that we do want some contacts between static/kinematic bodies to be computed by the narrow-phase, but ignored by the constraints solver. Maybe that could be useful for character controllers?

I would like to open the discussion on whether this can be useful and why. Please comment or add a :+1: bellow if you are interested in this feature, or :-1: if you think this is completely useless.

sebcrozet:
Fix #3
Fix #15 

This PR adds two traits `CollisionPairFilter` and `ProximityPairFilter` that allows user-defined logic for determining if two colliders/sensors are allowed to interact, as discussed in #3 

If these filters are set by the user (i.e. passed to the `PhysicsPipeline::step` method), then they replace the predefined filtering logic which consists of preventing contact computation between two non-dynamic bodies (the collision-groups on the other hands are still taken into account). That way, it is possible for the user to enable  contacts and proximity between two non-dynamic bodies to be computed, as suggested in #15 

Finally, the `CollisionPairFilter` returns a `SolverFlags` bitmask that will be used by the constraints solver to handle the contact manifolds for this collision pair. Currently, only one bit is implemented: `SolverFlags::COMPUTE_IMPULSES`. If this bit is set, then the constraints solver will compute impulses for the contact manifolds generated by this contact pair. If this bit is not set, these contact manifolds will be ignored by the constraints solver.

### DIFF
--- a/src/geometry/narrow_phase.rs
+++ b/src/geometry/narrow_phase.rs
@@ -405,7 +405,7 @@ impl NarrowPhase {
                 }
             } else {
                 if co1.solver_groups.test(co2.solver_groups) {
-                    SolverFlags::COMPUTE_FORCES
+                    SolverFlags::COMPUTE_IMPULSES
                 } else {
                     SolverFlags::empty()
                 }
@@ -454,7 +454,9 @@ impl NarrowPhase {
             for manifold in &mut inter.weight.manifolds {
                 let rb1 = &bodies[manifold.body_pair.body1];
                 let rb2 = &bodies[manifold.body_pair.body2];
-                if manifold.solver_flags.contains(SolverFlags::COMPUTE_FORCES)
+                if manifold
+                    .solver_flags
+                    .contains(SolverFlags::COMPUTE_IMPULSES)
                     && manifold.num_active_contacts() != 0
                     && (rb1.is_dynamic() || rb2.is_dynamic())
                     && (!rb1.is_dynamic() || !rb1.is_sleeping())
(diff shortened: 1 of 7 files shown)


sebcrozet:
It looks like this PR introduces a performance regression for the case where objects are all fallen asleep:

image
image

Investigating that now.

src/geometry/narrow_phase.rs
```diff
@@ -301,16 +303,35 @@ impl NarrowPhase {
             let rb1 = &bodies[co1.parent];
             let rb2 = &bodies[co2.parent];
 
-            if (rb1.is_sleeping() || rb1.is_static()) && (rb2.is_sleeping() || rb2.is_static()) {
-                // No need to update this contact because nothing moved.
+            if (rb1.is_sleeping() && rb2.is_static()) || (rb2.is_sleeping() && rb1.is_static()) {
+                // No need to update this proximity because nothing moved.
                 return;
             }
```
sebcrozet, line 309: This is the source of the performance regression when objects are resting.
With our change, two sleeping dynamic bodies will no longer be filtered-out by this test.

sebcrozet:
I fixed the regression:

image
image

This is ready to be merged now.

PR #57, opened 2020-11-05
Fixed asserts in Trimesh::new()

mrDIMAS:
Assert message tells "must contain at *least one*", but condition checks for 2 or more.

### DIFF
--- a/src/geometry/trimesh.rs
+++ b/src/geometry/trimesh.rs
@@ -19,11 +19,11 @@ impl Trimesh {
     /// Creates a new triangle mesh from a vertex buffer and an index buffer.
     pub fn new(vertices: Vec<Point<f32>>, indices: Vec<Point3<u32>>) -> Self {
         assert!(
-            vertices.len() > 1,
+            vertices.len() > 0,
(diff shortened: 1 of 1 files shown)


src/geometry/trimesh.rs
```diff
@@ -19,11 +19,11 @@ impl Trimesh {
     /// Creates a new triangle mesh from a vertex buffer and an index buffer.
     pub fn new(vertices: Vec<Point<f32>>, indices: Vec<Point3<u32>>) -> Self {
         assert!(
-            vertices.len() > 1,
+            vertices.len() > 0,
             "A triangle mesh must contain at least one point."
```
sebcrozet, line 23: Actually, the `> 1` test was correct, but the text was wrong.
It should be "A triangle mesh must contain at least three points.".

sebcrozet (changes requested):

sebcrozet:
Thank you for this PR! However I am closing it since it will become irrelevant once #79 lands (because the trimesh definition will be moved to another crate).

PR #68, opened 2020-11-24
Allow access to contact information

### LINKED ISSUE #56: Allow access to contact information
Right now, accessing to contact information (contact normals, forces, etc.) between colliders is nearly impossible. While the narrow-phase does give access to the contact graph, this graph cannot be queried because it requires some internal indices given to each colliders.

We should make it possible to retrieve contacts between colliders, using the collider handles instead of these internal indices.

sebcrozet:
Fix #56 

This PRs will allow the user to retrieve:
- All the contacts/proximities involving one specific collider handle.
- The contacts/proximity information for one specific pair of collider handles.

In order to achieve this, one significant internal change was made: the interaction graph IDs are no longer stored in the `Collider`. Instead, they are stored by the narrow-phase directly. This is great because it decouples things a little more. However this adds a few indirection at some places when browsing the contact graph in needed. So we need to see if this has any notable effect on performance.

### DIFF
--- a/examples3d/all_examples3.rs
+++ b/examples3d/all_examples3.rs
@@ -10,7 +10,6 @@ use inflector::Inflector;
 use rapier_testbed3d::Testbed;
 use std::cmp::Ordering;
 
-mod add_remove3;
 mod collision_groups3;
 mod compound3;
 mod damping3;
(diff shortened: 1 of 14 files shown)


sebcrozet:
Add good, no performance regression detected.

PR #70, opened 2020-11-30
Add a more convenient way of locking the rotations (or translations) of a rigid-body

sebcrozet:
Locking rotations isn't so easy right now: one needs to set the principal inertia of the rigid-body to zero **after** all the colliders has been attached to it. This can be annoying to do, especially when using an ECS like Bevy's (see https://github.com/dimforge/bevy_rapier/issues/9).

This PR introduces a few changes to make this much easier: it allows the user to specify, at construction time, if a rigid-body should ignore some contributions a collider would make to its mass properties. This allows making rotations along some coordinate axes, or all translations, behave in a kinematic way (i.e. only controlled by the user, immune to forces).

For convenience, there are the new `RigidBodyBuilder::lock_rotations` and `RigidBodyBuilder::lock_translations` that hide these details from the user, for simplicity. 

This includes a breaking chanhge because `RigidBodyBuilder::mass` takes an additional parameter now.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,12 @@
   their corresponding getters/setters. For example: `rb.linvel()`, `rb.set_linvel(vel, true)`.
 - Add `RigidBodyBuilder::sleeping(true)` to allow the creation of a rigid-body that is asleep
   at initialization-time.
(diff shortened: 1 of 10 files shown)


memoryruins:
Oh, this is a great addition!

PR #75, opened 2020-12-03
Replace the delta-position by the collider positions in the contact manifolds.

sebcrozet:
Currently, contact manifold contains two fields `delta1` and `delta2` which are the colliders' positions relative to their rigid-body. This piece of data already exist on colliders and is mostly constant. But we have them duplicated into the `ContactManifold` structure for performance reasons (to avoid lookups of collider data inside of the constraints solver).

This PR replaced these fields by `position1` and `position2` which are the colliders' world-space positions at the time the contacts were detected. That way, these field will also be useful for the end-user that would like to know the world-space location of contacts points and normals at the time of the contact. Without these the user could not retrieve these world-space data because the position of the colliders after the execution of the timestep would have already changed.

### DIFF
--- a/src/dynamics/rigid_body.rs
+++ b/src/dynamics/rigid_body.rs
@@ -381,6 +381,7 @@ impl RigidBody {
     }
 
     /// The world-space position of this rigid-body.
+    #[inline]
     pub fn position(&self) -> &Isometry<f32> {
         &self.position
     }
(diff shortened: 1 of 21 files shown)


sebcrozet:
Closing this PR as it won't be relevant any more once #79 lands.

PR #80, opened 2020-12-19
Extract Physics into module, make new public

rezural:
Mainly adds public constructors to PhysicsState & PhysicsSnapshot.

Also moves PhysicsState & PhysicsSnapshot  & PhysicsEvents to their own module.

I mainly need this for the run harness work (only the public new's). If you'd prefer to keep it within testbed, I can change this back.

### DIFF
--- a/src_testbed/lib.rs
+++ b/src_testbed/lib.rs
@@ -23,7 +23,7 @@ extern crate log;
 
 pub use crate::engine::GraphicsManager;
 pub use crate::plugin::TestbedPlugin;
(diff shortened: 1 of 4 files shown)


rezural:
send-bench-message fails on this, however the error indicates it is a curl error..?

sebcrozet:
Thank you for this PR. Yes, it is cleaner this way.
Don't wrorry about the `send-bench-message` CI failing. It doesn't work yet with PRs originating from forks.

PR #81, opened 2020-12-20
Introduce a run harness, to make running salva with rapier easier with headless.

rezural:


sebcrozet:
Thanks! Can you please rebase this PR on master? That way it won't show the diffs that were already included in your previous PR #80 (which I merged a few minutes ago).
Also, you may want to run a `cargo fmt` so that the `check-fmt` CI passes.

rezural:
OK, rebased, and formatted.

sebcrozet:
Looks good, thanks! I’m wondering if the testbed could store this harness struct internally and use its `step` function (instead of duplicating all the code calling callbacks and stepping the physics pipeline and plugins).

rezural:
I'll have a look at this in the next couple of days, or so.

I straight up copied the testbed code, without any testbed integration yet, but it definately makes sense to minimize the duplication.

PR #83, opened 2020-12-24
Harness testbed integration

rezural:
This moves some of the code from testbed, and delegates it to the harness.

There may be some mistakes, as the logic for benchmarking, & running in a window are delicate. 

If you want, I can look at moving the benchmark code to a separate crate, so it can be removed from testbed.

Cheers

### DIFF
--- a/examples2d/add_remove2.rs
+++ b/examples2d/add_remove2.rs
@@ -10,7 +10,7 @@ pub fn init_world(testbed: &mut Testbed) {
     let rad = 0.5;
 
     // Callback that will be executed on the main loop to handle proximities.
-    testbed.add_callback(move |window, graphics, physics, _, _| {
+    testbed.add_callback(move |mut window, mut graphics, physics, _, _| {
         let rigid_body = RigidBodyBuilder::new_dynamic()
             .translation(0.0, 10.0)
             .build();
@@ -20,10 +20,8 @@ pub fn init_world(testbed: &mut Testbed) {
             .colliders
             .insert(collider, handle, &mut physics.bodies);
 
-        if graphics.is_some() {
-            graphics
-                .unwrap()
-                .add(window.unwrap(), handle, &physics.bodies, &physics.colliders);
+        if let (Some(graphics), Some(window)) = (&mut graphics, &mut window) {
+            graphics.add(*window, handle, &physics.bodies, &physics.colliders);
         }
 
         let to_remove: Vec<_> = physics
@@ -37,10 +35,9 @@ pub fn init_world(testbed: &mut Testbed) {
                 .bodies
                 .remove(handle, &mut physics.colliders, &mut physics.joints);
 
-            // FIXME: need a way to access graphics & window in a loop
-            // if graphics.is_some() {
-            //     graphics.unwrap().remove_body_nodes(window.unwrap(), handle);
-            // }
+            if let (Some(graphics), Some(window)) = (&mut graphics, &mut window) {
+                graphics.remove_body_nodes(*window, handle);
+            }
         }
     });
 
--- a/examples2d/add_remove2.rs
+++ b/examples2d/add_remove2.rs
@@ -10,7 +10,7 @@ pub fn init_world(testbed: &mut Testbed) {
     let rad = 0.5;
 
     // Callback that will be executed on the main loop to handle proximities.
-    testbed.harness_mut().add_callback(move |physics, _, _| {
+    testbed.add_callback(move |window, graphics, physics, _, _| {
         let rigid_body = RigidBodyBuilder::new_dynamic()
             .translation(0.0, 10.0)
             .build();
@@ -20,8 +20,11 @@ pub fn init_world(testbed: &mut Testbed) {
             .colliders
             .insert(collider, handle, &mut physics.bodies);
 
-        // TODO: need a way to access graphics & window
-        // graphics.add(window, handle, &physics.bodies, &physics.colliders);
+        if graphics.is_some() {
+            graphics
+                .unwrap()
+                .add(window.unwrap(), handle, &physics.bodies, &physics.colliders);
+        }
 
         let to_remove: Vec<_> = physics
             .bodies
@@ -34,8 +37,10 @@ pub fn init_world(testbed: &mut Testbed) {
                 .bodies
                 .remove(handle, &mut physics.colliders, &mut physics.joints);
 
-            // TODO: need a way to access graphics & window
-            // graphics.remove_body_nodes(window, handle);
+            // FIXME: need a way to access graphics & window in a loop
+            // if graphics.is_some() {
+            //     graphics.unwrap().remove_body_nodes(window.unwrap(), handle);
+            // }
         }
     });
 
--- a/examples2d/add_remove2.rs
+++ b/examples2d/add_remove2.rs
@@ -19,7 +19,7 @@ pub fn init_world(testbed: &mut Testbed) {
         physics
             .colliders
             .insert(collider, handle, &mut physics.bodies);
-        
+
         // TODO: need a way to access graphics & window
         // graphics.add(window, handle, &physics.bodies, &physics.colliders);
 
@@ -33,7 +33,7 @@ pub fn init_world(testbed: &mut Testbed) {
             physics
                 .bodies
                 .remove(handle, &mut physics.colliders, &mut physics.joints);
-        
+
             // TODO: need a way to access graphics & window
             // graphics.remove_body_nodes(window, handle);
         }
--- a/examples2d/add_remove2.rs
+++ b/examples2d/add_remove2.rs
@@ -10,7 +10,7 @@ pub fn init_world(testbed: &mut Testbed) {
     let rad = 0.5;
 
     // Callback that will be executed on the main loop to handle proximities.
-    testbed.add_callback(move |window, physics, _, graphics, _| {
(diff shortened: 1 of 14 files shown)


rezural:
your thoughts on this are welcome. It's a fairly large change, which came out organically.

Feel free to be honest. I can look into chunking this into a few PR's

This also may be easier squashed, as some of the commits probably had errors.

src_testbed/testbed.rs
```diff
@@ -1172,9 +1123,10 @@ impl State for Testbed {
                         }
 
                         self.set_world(w.3, w.4, w.5);
-                        self.physics.broad_phase = w.1;
-                        self.physics.narrow_phase = w.2;
-                        self.state.timestep_id = w.0;
+                        self.harness.physics.broad_phase = w.1;
+                        self.harness.physics.narrow_phase = w.2;
+                        //FIXME: not completely sure this is valid
+                        self.harness.state.timestep_id = w.0;
```
sebcrozet, line 1129:
```suggestion
                        self.harness.state.timestep_id = w.0;
```
This should be valid, yes.

src_testbed/testbed.rs
```diff
@@ -1349,67 +1267,62 @@ impl State for Testbed {
...
-
                 for plugin in &mut self.plugins {
-                    plugin.run_callbacks(window, &mut self.physics, self.time)
+                    {
+                        plugin.run_callbacks(
+                            window,
+                            &mut self.harness.physics,
+                            &self.harness.state,
+                        );
+                    }
```
sebcrozet, line 1301:
```suggestion
                    plugin.run_callbacks(
                        window,
                        &mut self.harness.physics,
                        &self.harness.state,
                    );
```
Looks like there are some useless braces?

src_testbed/testbed.rs
```diff
@@ -419,6 +382,7 @@ impl Testbed {
     }
 
     fn clear(&mut self, window: &mut Window) {
+        //FIXME: do we need to do this still, after moving to harness code?
         self.callbacks.clear();
```
sebcrozet, line 386: Now that the harness is responsible for running the callbacks, all the callbacks-related code should be removed from the Testbed (and it turns out that the callbacks stored in the testbed are never called now anyways).

src_testbed/testbed.rs
```diff
@@ -464,6 +429,7 @@ impl Testbed {
             }
         }
 
+        //FIXME: move this to dedicated benchmarking code
```
sebcrozet, line 432:
```suggestion
        // TODO: move this to dedicated benchmarking code
```
Yeah, this should at least be in a separate function, on a separate module. But this is not important for this PR.

sebcrozet (changes requested):
Thank you for this PR! This looks all good besides a few details bellow.

src_testbed/testbed.rs
rezural: I think this will require either:

* updating the examples code to run callbacks on the harness
* having a forwarding call on testbed to harness

I image the first is preferred, I will look into this.

rezural: OK, looking now:

There is no way for callbacks to access the Window object through the harness. I'm not sure if this is an issue (probably a rare use case), but it may be..?

I will remove the callback code for now, perhaps it can be moved to a CallbackPlugin (that code still works I think), for use cases where it is needed.

rezural:
So!

In addition to your comments, I have:
* added harness_mut() to testbed.
* updated the examples to support this
* disabled graphics related code  in the examples (with TODO's) until some mechanism of handling window&graphics code from the harness, or the testbed is added back in


So, this is not strictly ready for prime time, but your feedback is probably best for deciding how to handle window & graphics aware code.

Thanks

rezural:
My first thought is to add a graphics aware callback to the harness, that then gets called via the testbed.

the method signature would perhaps be:
```rust
type Callbacks = Vec<Box<dyn FnMut(Option<&Window>, Option<&Graphics>, &mut PhysicsState, &PhysicsEvents, &RunState)>>;
```
note the Option, so that callback code can gracefully handle running with graphics and without, and not duplicating code.

sebcrozet:
Thank you for the changes!

Regarding callbacks, I think you could just add the `Option<&mut Window>` parameter to the existing harness callbacks instead of defining a new `GraphicsCallback`.

Regarding the examples, the `.harness_mut().add_graphics_callback` looks a bit more verbose than it could be. Actually, we just keep a `add_callback` method into the testbed:

```rust
impl Testbed {
        pub fn add_callback<
        F: FnMut(Option<&mut Window>, &mut PhysicsState, &PhysicsEvents, &HarnessState, f32) + 'static,
    > {
        self.harness.add_callback(callback);
    }
}
```

that way our examples don't even have to be aware of the harness inside of the testbed.

rezural:
Right,

Agreed on both. I'll have this up later today or tomorrow.

Cheers

examples2d/add_remove2.rs
```diff
@@ -19,7 +19,12 @@ pub fn init_world(testbed: &mut Testbed) {
         physics
             .colliders
             .insert(collider, handle, &mut physics.bodies);
-        graphics.add(window, handle, &physics.bodies, &physics.colliders);
+
+        if graphics.is_some() {
+            graphics
+                .unwrap()
+                .add(window.unwrap(), handle, &physics.bodies, &physics.colliders);
+        }
```
sebcrozet, line 27: Something like this should work:

```suggestion
        if let (Some(graphics), Some(window)) = (&mut graphics, &mut window) {
            graphics.add(*window, handle, &physics.bodies, &physics.colliders);
        }
```
Matching on `&mut Option` will make sure that you don't move the `&mut T` out of `Option<&mut T>`.

examples2d/add_remove2.rs
```diff
@@ -10,7 +10,7 @@ pub fn init_world(testbed: &mut Testbed) {
     let rad = 0.5;
 
     // Callback that will be executed on the main loop to handle proximities.
-    testbed.add_callback(move |window, physics, _, graphics, _| {
+    testbed.add_callback(move |window, graphics, physics, _, _| {
```
sebcrozet, line 13: This will be needed by the other suggestions that do `&mut window`, and `&mut graphics`:

```suggestion
    testbed.add_callback(move |mut window, mut graphics, physics, _, _| {
```

examples2d/add_remove2.rs
```diff
@@ -31,7 +36,11 @@ pub fn init_world(testbed: &mut Testbed) {
             physics
                 .bodies
                 .remove(handle, &mut physics.colliders, &mut physics.joints);
-            graphics.remove_body_nodes(window, handle);
+
+            // FIXME: need a way to access graphics & window in a loop
+            // if graphics.is_some() {
+            //     graphics.unwrap().remove_body_nodes(window.unwrap(), handle);
+            // }
```
sebcrozet, line 43: We can do the same as mentioned above:

```suggestion
            if let (Some(graphics), Some(window)) = (&mut graphics, &mut window) {
                 graphics.remove_body_nodes(*window, handle);
            }
```

examples2d/sensor2.rs
```diff
@@ -80,10 +80,12 @@ pub fn init_world(testbed: &mut Testbed) {
...
             if parent_handle1 != ground_handle && parent_handle1 != sensor_handle {
-                graphics.set_body_color(parent_handle1, color);
+                // FIXME: need a way to access graphics & window in a loop
+                // graphics.set_body_color(parent_handle1, color);
             }
             if parent_handle2 != ground_handle && parent_handle2 != sensor_handle {
-                graphics.set_body_color(parent_handle2, color);
+                // FIXME: need a way to access graphics & window in a loop
+                // graphics.set_body_color(parent_handle2, color);
             }
```
sebcrozet, line 89: Here we can do:

```rust
if let Some(graphics) = &mut graphics {
    if ... {
    }
    if ... {
    }
}
```

examples3d/debug_add_remove_collider3.rs
```diff
@@ -43,7 +43,15 @@ pub fn init_world(testbed: &mut Testbed) {
...
-        graphics.add_collider(window, ground_collider_handle, &physics.colliders);
+
+        // TODO: need a way to access graphics & window
+        if graphics.is_some() {
+            graphics.unwrap().add_collider(
+                window.unwrap(),
+                ground_collider_handle,
+                &physics.colliders,
+            );
+        }
```
sebcrozet, line 54:
```suggestion
        if let (Some(graphics), Some(window)) = (graphics, window) {
            graphics.add_collider(
                window,
                ground_collider_handle,
                &physics.colliders,
            );
        }
```

examples2d/add_remove2.rs
rezural: Right,

I think I get this. Regardless, I have changed all the callback related example code to reflect these suggestions.

Thank you, this saved me hours of hair pulling.

Thanks

nhemsley:
OK,

this last commit compiles (I will check that all the examples still work as well, by tomorrow).

Thanks

examples2d/sensor2.rs
nhemsley: OK, looks good.

rezural:
OK, these changes are made, 

thankyou for the newbiew help with rust! Learning every day.

Cheers

sebcrozet:
All good, thank you!

PR #88, opened 2021-01-21
Small refactor of IntegrationParameters

emilk:
I found the use of `IntegrationParameters::new(…)` hard to read (difficult to match value to parameter name), so I replaced it with `Self { … }` and immediately noticed that the documentation for `allowed_linear_error` was wrong.

It would be nice to deprecate `IntegrationParameters::new` (so many arguments are bound to lead to bugs), but I didn't want to break old code.

It would also be nice to have all members public, as users could then simply write `IntegrationParameters { velocity_iterations: 10, ..Default::default() }`, but then we run the risk of `dt != 1.0 / inv_dt`. I suspect we could remove `inv_dt` and compute it once per constraint per frame without a noticeable impact in performance (or compute it once and pass it as an additional parameter to the `generate` functions).

If deprecating `IntegrationParameters::new` and/or removing `inv_dt` sounds good I can do so in this or another PR.


sebcrozet:
Thank you for this PR! I agree that being able to do `IntegrationParameters { velocity_iterations: 10, ..Default::default() }` is much nicer than the current `::new`. I think this it is OK for this PR to `#[deprecate]` the `IntegrationParameters::new` function. Since it is just a deprecation (for now) it should not break any downstream crate (unless they have `#![deny(warnings)]` or something similar).

If all the fields of `IntegrationParameters` become public, then I think we should remove `inv_dt` immediately too. Otherwise we may end up introducing unexpected downstream bugs if one is not careful enough to update both `dt` and `inv_dt` at the same time. Computing `inv_dt` once on each `::generate` function should indeed have no impact on performances.

sebcrozet:
All good, thanks!

PR #98, opened 2021-02-01
Split bouncy and resting contacts

emilk:
Various experiments. Needs rebasing on `master` now that rapier 0.5 has been released. Or more likely I'll reimplement the ideas in a more clean manner.

The goal of this work was to get energy conservation for bouncing spheres when restitution=1, not because that is something important in itself, but because failing to conserve energy is symptomatic of larger problems.

The longer goal is that by splitting resting and bouncing contacts we can start applying corrective impulses in the velocity solver, which will mean we won't need the positional/projection solver. Position iterations is always a bad idea, so getting rapier to a point where it is optional would be very good.

### DIFF
--- a/src/dynamics/rigid_body.rs
+++ b/src/dynamics/rigid_body.rs
@@ -125,12 +125,27 @@ impl RigidBody {
         self.active_set_timestamp = 0;
     }
 
-    pub(crate) fn integrate_accelerations(&mut self, dt: f32, gravity: Vector<f32>) {
-        if self.mass_properties.inv_mass != 0.0 {
-            self.linvel += (gravity + self.linacc) * dt;
-            self.angvel += self.angacc * dt;
+    // pub(crate) fn integrate_accelerations(&mut self, dt: f32, gravity: Vector<f32>) {
+    //     if self.mass_properties.inv_mass != 0.0 {
+    //         self.linvel += (gravity + self.linacc) * dt;
+    //         self.angvel += self.angacc * dt;
 
-            // Reset the accelerations.
+    //         // Reset the accelerations.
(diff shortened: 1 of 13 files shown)


sebcrozet:
Hi! I see that the "bouncing" status is given to contact manifolds (instead of contact themselves). Is this on purpose? I feel like this should a be per-contact attributes. For example if we have a long capsule (initialized making a 45 degrees angle with the floor) with one end touching the floor. Due to gravity the second ends touch the floor after some time. I would expect this second contact point (which would be part of the same contact manifold as the first contact point) to result in a bounce.

And if it's a per-contact attribute, I would think that a "bouncing" contact would be a contact where its cached impulse is zero (meaning that it was not an active contact during the last frame) instead of maintaining a boolean.

emilk:
@sebcrozet good feedback!

I considered having the split per-contact instead of per-manifold. While experimenting, doing the decision per manifold turned out easier because it allowed me to disable the position solver for bouncing contacts. But now that rapier 0.5 does position solving based on each contact point (instead of the whole manifold), it would be better to use per-contact bounce/rest split again.

The `bool` was also a bit of a hack just to test out the idea. I did consider using `impulse != 0`, but there is a subtle failure-case there, where a contact can be resting yet have a zero impulse if the other contacts in the manifold "carries" it fully. But I doubt it will cause problems in practice, but I just wanted to exclude the possibility for this experiment (especially since I don't fully grok all parts of rapier yet).

In summary: in my next attempt at this I will use per-contact split with `impulse != 0` as per your suggestion :)

emilk:
Reimplemented in better PRs

Issue #97, opened 2021-01-26
[BUG] Energy is not conserved for perfectly elastic collisions


emilk:
**The test**: ball dropped on a static box, both with restitution=1.
**Expected**: the ball bounce back up to the same height.
**Actual**: the ball bounces higher and higher.

The good news is I have a fix that works locally. Just need to clean it up a bit :)

---

Latest `master`:

<img width="326" alt="Screen Shot 2021-01-27 at 10 12 52" src="https://user-images.githubusercontent.com/1148717/105969646-c44afe80-6088-11eb-8397-6b69c9c723f8.png">

With my changes:

<img width="321" alt="Screen Shot 2021-01-27 at 10 11 46" src="https://user-images.githubusercontent.com/1148717/105969665-ca40df80-6088-11eb-84d8-948faeffa1d9.png">

Energy is conserved to within 1e-5 per bounce.

PR #100, opened 2021-02-03
Switch from Explicit Euler integration to Symplectic Euler

emilk:
Explicit Euler tends to add energy to the system,
which in the worst case can lead to divergence and explosion.

Symplectic Euler is energy conserving.

A simple test is ball bouncing on a fixed plane, both with restitution=1 (https://github.com/dimforge/rapier/issues/97).
With the old euler integration, the ball would quickly bounce higher and higher.
With the new symplectic euler, the ball slowly loses energy instead, which is much better.

(This energy loss is because gravity is applied before the velocity iterations.
In a later PR I will instead make accelerations from gravity and other external forces
be part of the velocity solver. This should eliminate this energy leakage.)

### DIFF
--- a/src/dynamics/solver/parallel_island_solver.rs
+++ b/src/dynamics/solver/parallel_island_solver.rs
@@ -217,6 +217,7 @@ impl ParallelIslandSolver {
                     std::mem::transmute(parallel_joint_constraints.load(Ordering::Relaxed))
                 };
 
+                // Symplectic Euler: initialize contraints with the *old* positions.
                 enable_flush_to_zero!(); // Ensure this is enabled on each thread.
                 parallel_contact_constraints.fill_constraints(&thread, params, bodies, manifolds);
                 parallel_joint_constraints.fill_constraints(&thread, params, bodies, joints);
@@ -229,6 +230,16 @@ impl ParallelIslandSolver {
                     parallel_joint_constraints.constraint_descs.len(),
                 );
 
+                // Symplectic Euler: move bodies using the *old* velocities.
+                concurrent_loop! {
+                    let batch_size = thread.batch_size;
+                    for handle in active_bodies[thread.body_integration_index, thread.num_integrated_bodies] {
+                        let rb = &mut bodies[handle.0];
+                        rb.integrate(params.dt);
+                        positions[rb.active_set_offset] = rb.position;
+                    }
+                }
+
                 ParallelVelocitySolver::solve(
                         &thread,
                         params,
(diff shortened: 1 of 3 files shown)


src/dynamics/solver/parallel_island_solver.rs
```diff
@@ -229,6 +230,16 @@ impl ParallelIslandSolver {
...
+                // Symplectic Euler: move bodies using the *old* velocities.
+                concurrent_loop! {
+                    let batch_size = thread.batch_size;
+                    for handle in active_bodies[thread.body_integration_index, thread.num_integrated_bodies] {
+                        let rb = &mut bodies[handle.0];
+                        rb.integrate(params.dt);
+                        positions[rb.active_set_offset] = rb.position;
+                    }
+                }
+
```
emilk, line 242: I am kind of winging it here with the concurrent version.

The `concurrent_loop` macro is a bit... undocumented :P

emilk:
Is there a test that checks if `ParallelIslandSolver` act the same as `IslandSolver` ?

sebcrozet:
Hi!
I am not sure I understand the changes here. As far as I understand:
- Symplectic Euler updates velocities first, and use these new velocities to update the positions.
- Explicit Euler update positions with the old velocities.

So I believe that the existing code already implements Symplectic Euler, and that this PR is replacing it by Explicit Euler (the opposite from what the PR title claims).

> This energy loss is because gravity is applied before the velocity iterations.
In a later PR I will instead make accelerations from gravity and other external forces
be part of the velocity solver. This should eliminate this energy leakage.

Now that you mention it, I believe this is the origin of the restitution problem: the contributions from external forces should not be included in the restitution computation. Fixing this while keeping the current integration scheme appears to fix the increase of energy.

sebcrozet:
> Is there a test that checks if `ParallelIslandSolver` produces the same results as `IslandSolver` ?

The `ParallelIslandSolver` will generally not produce the same results as `IslandSolver`: both solve constraints in a different order. The `ParallelIslandSolver` performs a complete graph coloring to extract independent contacts per-island. The `IslandSolver` performs a more local version of the graph coloring (because it knows only 4 constraints will be solved in parallel at any given times thanks to SIMD).

emilk:
Yeah, you are right about the integrator... I misremembered. Still I needed this change in my other PR to get perfect energy conservation, but I'm no longer sure why. I will need to investigate further.

emilk:
> Now that you mention it, I believe this is the origin of the restitution problem: the contributions from external forces should not be included in the restitution computation. Fixing this while keeping the current integration scheme appears to fix the increase of energy.

I can make a PR where forces are applied in the velocity solver instead of in a the current `integrate_accelerations`.

PR #102, opened 2021-02-03
Apply accelerations during velocity solver

emilk:
Closes https://github.com/dimforge/rapier/issues/97

A ball bouncing on a plane with restitution=1 now perfectly conserves energy to within machine precision (within ~1e-7 per bounce).

Instead of applying accelerations from gravity and external forces
as a separate step, this PR switches to applying them in the velocity solver.

### DIFF
--- a/src/dynamics/solver/island_solver.rs
+++ b/src/dynamics/solver/island_solver.rs
@@ -35,7 +35,9 @@ impl IslandSolver {
         joints: &mut [JointGraphEdge],
         joint_indices: &[JointIndex],
     ) {
-        if manifold_indices.len() != 0 || joint_indices.len() != 0 {
+        let has_constraints = manifold_indices.len() != 0 || joint_indices.len() != 0;
+
+        if has_constraints {
             counters.solver.velocity_assembly_time.resume();
             self.contact_constraints
                 .init(island_id, params, bodies, manifolds, manifold_indices);
@@ -54,13 +56,13 @@ impl IslandSolver {
                 &mut self.joint_constraints.velocity_constraints,
             );
             counters.solver.velocity_resolution_time.pause();
-        }
 
-        counters.solver.velocity_update_time.resume();
-        bodies.foreach_active_island_body_mut_internal(island_id, |_, rb| rb.integrate(params.dt));
-        counters.solver.velocity_update_time.pause();
+            counters.solver.velocity_update_time.resume();
+            bodies.foreach_active_island_body_mut_internal(island_id, |_, rb| {
+                rb.integrate(params.dt)
+            });
+            counters.solver.velocity_update_time.pause();
 
-        if manifold_indices.len() != 0 || joint_indices.len() != 0 {
             counters.solver.position_resolution_time.resume();
             self.position_solver.solve(
                 island_id,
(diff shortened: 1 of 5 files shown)


src/dynamics/solver/island_solver.rs
```diff
@@ -70,6 +72,14 @@ impl IslandSolver {
                 &self.joint_constraints.position_constraints,
             );
             counters.solver.position_resolution_time.pause();
+        } else {
+            counters.solver.velocity_update_time.resume();
+            bodies.foreach_active_island_body_mut_internal(island_id, |_, rb| {
+                // Since we didn't run the velocity solver we need to integrate the accelerations here
+                rb.integrate_accelerations(params.dt);
```
emilk, line 79: It seems like there is nothing like this `else` in the parallel solver. Is everything part of an island in the parallel case?

sebcrozet: For both the parallel and serial solvers every non-sleeping body is part of an island. The `else` doesn't exist in the parallel solver because it does not have the `if manifold_indices.len() != 0 || joint_indices.len() != 0 {` either. But I guess it could. This test is just here to avoid entering in code that will just loop on empty arrays.

sebcrozet (changes requested):
Thank you for this PR!

There is one probalem with the angular acceleration. It's not documented (and the struct name `DeltaVel` is misleading), but `mj_lambda` contains the linear velocity delta but not the angular velocity delta. It contains the angular velocity delta multiplied by the square root of the inertia tensor. So it should be initialized as something like this:

 ```rust
 dvel.angular += rb.effective_world_inv_inertia_sqrt.inverse_unchecked() * rb.angacc * params.dt;`
```
The use of `.inverse_unchecked()` sounds unfortunate performance-wise and isn't great as it won't work with rigid-bodies with some locked rotation axes.
Perhaps we should not store `angacc` and store the torque applied to the rigid-body instead. That way the initialization of `angular` becomes:

```rust
 dvel.angular += rb.effective_world_inv_inertia_sqrt * rb.torque * params.dt;`
```

An alternative that does not require multiplying by the square-root inertia is to take the acceleration into account at the constraints level, when computing `rhs`. Though that would be more cumbersome because that would have to be taken into account into all the constraints, including joints.

On a separate note, there is something strange with one of the benchmark runs:
image

It appears that this PR makes the convex-polyhedron demo fall asleep sooner than before (see how the simulation time for the grey curve drops suddenly? That's when everything starts sleeping). This is a bit surprising considering this is a demo without restitution so it should not be affected by your change. But I guess this _can_ happen since the simulation won't be exactly the same as before since the acceleration term isn't applied at the same place as before… So maybe there is nothing to be concerned about here.

sebcrozet:
To clarify, the reason why `DeltaVel::angular` does not contain the delta angular velocity is because the premultiplication by `effective_world_inv_inertia_sqrt` allow us to avoid storing the rigid-bodies' angular inertia into the constraint structures. This saves a few computations during constraints resolution and keeps the constraints structures smaller.

emilk:
I think storing `force+torque` instead of `linacc+angacc` per body makes sense.

emilk:
Rebased upon https://github.com/dimforge/rapier/pull/108 and with the correct (I think) `dvel.angular`

sebcrozet (approved):

sebcrozet:
All good, thanks!

PR #103, opened 2021-02-04
Add functions for kinetic and potential energy of a RigidBody

emilk:
This was helpful when diagnosing https://github.com/dimforge/rapier/issues/97

I figure it is up to the caller to include or ignore non-dynamic bodies.


sebcrozet:
Thanks!

PR #104, opened 2021-02-04
Make clippy a bit happier

emilk:
I did a quick pass fixing some things that clippy complained about. Nothing major. I there is anything you disagree with I am happy to revert.

### DIFF
--- a/src/dynamics/solver/joint_constraint/revolute_position_constraint.rs
+++ b/src/dynamics/solver/joint_constraint/revolute_position_constraint.rs
@@ -56,7 +56,7 @@ impl RevolutePositionConstraint {
         let axis1 = position1 * self.local_axis1;
         let axis2 = position2 * self.local_axis2;
         let delta_rot =
-            Rotation::rotation_between_axis(&axis1, &axis2).unwrap_or(Rotation::identity());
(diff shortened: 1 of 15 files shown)


src/dynamics/solver/joint_constraint/revolute_position_constraint.rs
```diff
@@ -56,7 +56,7 @@ impl RevolutePositionConstraint {
         let axis1 = position1 * self.local_axis1;
         let axis2 = position2 * self.local_axis2;
         let delta_rot =
-            Rotation::rotation_between_axis(&axis1, &axis2).unwrap_or(Rotation::identity());
+            Rotation::rotation_between_axis(&axis1, &axis2).unwrap_or_else(Rotation::identity);
```
emilk, line 59: maybe `.unwrap_or_default()` would be even better?

sebcrozet: I like the more explicit `.unwrap_or_else(Rotation::identity)` better. The docs about `Default` states:

> Sometimes, you want to fall back to some kind of default value, and don't particularly care what it is [...]

Since we do care about the fact we get the identity, it feels like `Default` isn't appropriate.

sebcrozet:
Everything is OK. Thanks!

PR #108, opened 2021-02-08
Replace linacc/angacc with force/torque inside of RigidBody

emilk:
I also improved the documentation for the various force/impulse applying functions.


src/dynamics/rigid_body.rs
```diff
@@ -467,25 +472,30 @@ impl RigidBody {
             }
         }
     }
+}
 
-    /*
-     * Application of forces/impulses.
-     */
+/// ## Applying forces and torques
+impl RigidBody {
```
emilk, line 478: This sections the documentation, which I think is nice.

emilk:
As discussed in https://github.com/dimforge/rapier/pull/102

PR #114, opened 2021-02-15
Split bouncy and resting contacts (take 2)

emilk:
If a contact is new (previous impulse = 0), then we treat it as bouncy (respecting restitution). If the contact is old we treat it as resting.

This allows us to remove the `restitution_velocity_threshold` parameter.

Special cases:
* `restitution <= 0` is always treated as resting
* `restitution >= 1` is always treated as bouncing

# Test

A 0.9 restitution ball bouncing on ground with the same (very high) restitution.

## Before

Previously an arbitrary `restitution_velocity_threshold` cutoff led to an abrupt end to bounces:

previous

Screen Shot 2021-02-15 at 21 22 01

## After

Bounces keep going up to nyquist limit, into a smooth transition to rest:

Kapture 2021-02-15 at 21 28 11

Screen Shot 2021-02-15 at 21 28 50

PR #116, opened 2021-02-16
Optional violation correction in velocity solver

emilk:
This PR introduces a restorative impulse in the velocity solver that aims to resolve the constraint violation. The default strength of this is zero, i.e. it is disabled.

This PR also introduces a multiplier for the velocity correction of the velocity solver. The default is `1` (keeping the old behavior).

## Other changes / improvements

I fixed a bug in the prismatic limit joint that used to cause violations of the conservation of momentum.

I also added support for having the prismatic joints range be zero or negative (i.e. activating both sides of the prismatic limits).

## Benefits

Some of the benefits of having the violation resolution in the velocity solver include:

* No need for the positional solver
  * Saves some CPU time
  * Saves us from potential energy violations
* Faster violation resolution
  * Since we generally run the velocity solver for more iterations we have more iterations to converge on a solution
* Better coupling between velocity and position violations

## Downsides

Initial overlaps between colliders introduces a separating velocity, which makes the objects shoot away from each other rather than slowly separate.

# Progress

There are five constraints in Rapier (contact, ball, fixed, prismatic and revolute). Each has a scalar and simd (wide) version. Each has a dynamic-dynamic and dynamic-static ("ground") version. Each has a 2d and 3d version. So there are 5x2x2x2 = 40 things to update in this PR.

* [x] contact
* [x] ball
* [x] fixed
* [x] prismatic
* [x] revolute
* [x] Rebase on https://github.com/dimforge/rapier/pull/114
* [x] Rebase on https://github.com/dimforge/rapier/pull/119
* [x] Self review
* [x] Manual testing of the narrow versions (though not of `flipped` vs `!flipped`)
* [ ] Systematic testing

# Open questions

**For contacts**: How should we handle `allowed_linear_error`? It is currently pretty large (5 mm), and we also have `prediction_distance` (at 2mm).

## Testing

I've found these parameters work well:

```
max_position_iterations: 0,
warmstart_coeff: 0.5,
velocity_solve_fraction: 0.95,
velocity_based_erp: 0.20,
```

(and the rest left at default)

It would be nice to be able to test this in some automated fashion. That is, run the test barrage with these params and see that the tests works *approximately* the same as they do with the default parameters.

...though, is there even a test barrage? There is `src_testbed`, but I haven't found any docs or explanation for it. Seems to be only for manual testing?

# Results

In the subsequent test I compare the default parameters with the one described above.

## Contacts

When stirring around in a bowl the position solver will often damp out the movement, and the objects will tunnel though each other:

https://user-images.githubusercontent.com/1148717/108525718-b3308e80-72d0-11eb-870a-3c5076428c78.mp4

With the new velocity-based solver objects are properly displaced and move to make way for the object being dragged:

https://user-images.githubusercontent.com/1148717/108525782-c2afd780-72d0-11eb-8c8f-c79c991e1a4e.mp4

## Fixjoint resolution

In this test I have created a chain of bodies attached by fixjoints, with huge initial violations.

### Before

The position solver is slow to converge towards equilibrium, and the bodies falls asleep before reaching it (I need to poke the root body to wake up the chain):

rapier-fixed-before

### After

rapier-fixed-after

We see a much faster convergence. An analogy to machine learning comes to mind: while the position solver does gradient decent (with lr=0.2), the new solver does gradient decent with momentum.

## Revolute resolution

In this test a series of boxes are connected with revolute joints (the root is connected to the world).

### Before
The position solver does not seem to at all converge on a solution. I'm not sure if this is a bug or not:

revolute-rapier

### After
The new velocity solver converges after some swaying:

revolute-velocity-2

increasing the number of velocity iteration makes it converge faster (as expected).

### DIFF
--- a/src/dynamics/solver/joint_constraint/ball_velocity_constraint.rs
+++ b/src/dynamics/solver/joint_constraint/ball_velocity_constraint.rs
@@ -40,17 +40,21 @@ impl BallVelocityConstraint {
         rb2: &RigidBody,
         joint: &BallJoint,
     ) -> Self {
-        let anchor1 = rb1.position * joint.local_anchor1 - rb1.world_com;
-        let anchor2 = rb2.position * joint.local_anchor2 - rb2.world_com;
+        let anchor_world1 = rb1.position * joint.local_anchor1;
+        let anchor_world2 = rb2.position * joint.local_anchor2;
+        let anchor1 = anchor_world1 - rb1.world_com;
+        let anchor2 = anchor_world2 - rb2.world_com;
 
         let vel1 = rb1.linvel + rb1.angvel.gcross(anchor1);
         let vel2 = rb2.linvel + rb2.angvel.gcross(anchor2);
         let im1 = rb1.effective_inv_mass;
         let im2 = rb2.effective_inv_mass;
 
-        let rhs = -(vel1 - vel2);
-        let lhs;
+        let mut rhs = params.velocity_solve_fraction * (vel2 - vel1);
 
+        rhs += params.velocity_based_erp * params.inv_dt() * (anchor_world2 - anchor_world1);
+
+        let lhs;
         let cmat1 = anchor1.gcross_matrix();
         let cmat2 = anchor2.gcross_matrix();
 
@@ -271,22 +275,27 @@ impl BallVelocityGroundConstraint {
         joint: &BallJoint,
         flipped: bool,
     ) -> Self {
-        let (anchor1, anchor2) = if flipped {
+        let (anchor_world1, anchor_world2) = if flipped {
             (
-                rb1.position * joint.local_anchor2 - rb1.world_com,
-                rb2.position * joint.local_anchor1 - rb2.world_com,
+                rb1.position * joint.local_anchor2,
+                rb2.position * joint.local_anchor1,
             )
         } else {
             (
-                rb1.position * joint.local_anchor1 - rb1.world_com,
-                rb2.position * joint.local_anchor2 - rb2.world_com,
+                rb1.position * joint.local_anchor1,
+                rb2.position * joint.local_anchor2,
             )
         };
+
+        let anchor1 = anchor_world1 - rb1.world_com;
+        let anchor2 = anchor_world2 - rb2.world_com;
 
         let im2 = rb2.effective_inv_mass;
         let vel1 = rb1.linvel + rb1.angvel.gcross(anchor1);
         let vel2 = rb2.linvel + rb2.angvel.gcross(anchor2);
-        let rhs = vel2 - vel1;
+        let mut rhs = params.velocity_solve_fraction * (vel2 - vel1);
+
+        rhs += params.velocity_based_erp * params.inv_dt() * (anchor_world2 - anchor_world1);
 
         let cmat2 = anchor2.gcross_matrix();
 
--- a/src/dynamics/solver/joint_constraint/ball_velocity_constraint.rs
+++ b/src/dynamics/solver/joint_constraint/ball_velocity_constraint.rs
@@ -50,9 +50,8 @@ impl BallVelocityConstraint {
         let im1 = rb1.effective_inv_mass;
         let im2 = rb2.effective_inv_mass;
 
-        let mut rhs = params.velocity_solve_fraction * (vel2 - vel1);
-
-        rhs += params.velocity_based_erp * params.inv_dt() * (anchor_world2 - anchor_world1);
+        let rhs = (vel2 - vel1) * params.velocity_solve_fraction
+            + (anchor_world2 - anchor_world1) * params.velocity_based_erp_inv_dt();
 
         let lhs;
         let cmat1 = anchor1.gcross_matrix();
@@ -293,9 +292,9 @@ impl BallVelocityGroundConstraint {
         let im2 = rb2.effective_inv_mass;
         let vel1 = rb1.linvel + rb1.angvel.gcross(anchor1);
         let vel2 = rb2.linvel + rb2.angvel.gcross(anchor2);
-        let mut rhs = params.velocity_solve_fraction * (vel2 - vel1);
 
-        rhs += params.velocity_based_erp * params.inv_dt() * (anchor_world2 - anchor_world1);
+        let rhs = (vel2 - vel1) * params.velocity_solve_fraction
+            + (anchor_world2 - anchor_world1) * params.velocity_based_erp_inv_dt();
 
         let cmat2 = anchor2.gcross_matrix();
 
--- a/src/dynamics/solver/joint_constraint/prismatic_velocity_constraint.rs
+++ b/src/dynamics/solver/joint_constraint/prismatic_velocity_constraint.rs
@@ -48,12 +48,13 @@ pub(crate) struct PrismaticVelocityConstraint {
     motor_inv_lhs: Real,
     motor_max_impulse: Real,
 
+    limits_active: bool,
     limits_impulse: Real,
     /// World-coordinate direction of the limit force on rb2.
     /// The force direction on rb1 is opposite (Newton's third law)..
     limits_forcedir2: Vector<Real>,
     limits_rhs: Real,
-    limits_inv_lhs: Option<Real>,
+    limits_inv_lhs: Real,
     /// min/max applied impulse due to limits
     limits_impulse_limits: (Real, Real),
 
@@ -207,13 +208,12 @@ impl PrismaticVelocityConstraint {
             joint.motor_max_impulse,
         );
 
-        /*
-         * Setup limit constraint.
-         */
+        // Setup limit constraint.
+        let mut limits_active = false;
         let limits_forcedir2 = axis2.into_inner(); // hopefully axis1 is colinear with axis2
         let mut limits_rhs = 0.0;
         let mut limits_impulse = 0.0;
-        let mut limits_inv_lhs = None;
+        let mut limits_inv_lhs = 0.0;
         let mut limits_impulse_limits = (0.0, 0.0);
 
         if joint.limits_enabled {
@@ -233,7 +233,8 @@ impl PrismaticVelocityConstraint {
                 limits_impulse_limits.0 = -Real::INFINITY;
             }
 
-            if min_enabled || max_enabled {
+            limits_active = min_enabled || max_enabled;
+            if limits_active {
                 limits_rhs = (anchor_linvel2.dot(&axis2) - anchor_linvel1.dot(&axis1))
                     * params.velocity_solve_fraction;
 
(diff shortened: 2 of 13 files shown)


emilk:
There will be some (minor) merge conflicts with https://github.com/dimforge/rapier/pull/114, so it would be good to merge that first.

src/dynamics/solver/joint_constraint/prismatic_velocity_constraint.rs
```diff
@@ -141,13 +141,45 @@ impl PrismaticVelocityConstraint {
...
+                *axis1 * ((limit_err - max_limit).max(0.0) - (min_limit - limit_err).max(0.0));
+
+            #[cfg(feature = "dim2")]
+            {
+                rhs += Vector2::new(linear_err.x, ang_err.angle()) * velocity_based_erp_inv_dt;
+            }
+            #[cfg(feature = "dim3")]
+            {
+                let ang_err = ang_err.scaled_axis();
+                rhs += Vector5::new(linear_err.x, linear_err.y, ang_err.x, ang_err.y, ang_err.z)
```
emilk, line 179: I just realized we need `rhs` to be a `Vector6` here for limits to work

emilk: Another solution is to implement prismatic limits as a separate 1dof linear limit constraint

src/dynamics/solver/joint_constraint/revolute_velocity_constraint.rs
```diff
@@ -78,9 +78,30 @@ impl RevoluteVelocityConstraint {
...
+        if velocity_based_erp_inv_dt != 0.0 {
+            let lin_err = anchor2 - anchor1;
+
+            let axis1: na::Unit<na::Vector3<_>> = rb1.position * cparams.local_axis1;
+            let axis2: na::Unit<na::Vector3<_>> = rb2.position * cparams.local_axis2;
+            let ang_err =
+                Rotation::rotation_between_axis(&axis1, &axis2).unwrap_or_else(Rotation::identity);
+            let ang_err = ang_err.scaled_axis();
+
+            rhs += Vector5::new(lin_err.x, lin_err.y, lin_err.z, ang_err.x, ang_err.y)
```
emilk, line 102: This should also be a Vector6

src/dynamics/solver/joint_constraint/prismatic_velocity_constraint.rs
sebcrozet: Since the limit is an unilateral constraint, it is preferable to model it as a separate 1D constraint. Otherwise it won't interact well with the 5 other bilateral constraints. In particular, using a `Vector6` implies that the inverse Delassus matrix (`inv_lhs`) becomes 6x6 that either:
- Couples the joint limit constraint with the others, which is invalid if the limit isn't violated.
- Or not couple them (by setting all the off-diagonal elements of the 6-th row and column of the Delassus matrix), in which case the formulation is correct but this is equivalent to solve a 1D limit constraint and a 5D hard bilateral constraint in a Jacobi fashion (instead of a Gauss-Seidel) so this would probably reduce convergence rate.

emilk: I guess I should just use the existing `PrismaticVelocityGroundConstraint::limits_rhs`

sebcrozet: > I guess I should just use the existing PrismaticVelocityGroundConstraint::limits_rhs

Yes, this should do the trick.

emilk: I'll take this opportunity to also change `limits_forcedirs` to be just one vector, as the direction of forces on the two bodies should be opposite and equal

sebcrozet: > I'll take this opportunity to also change limits_forcedirs to be just one vector, as the direction of forces on the two bodies should be opposite and equal

If possible it would be best to keep that change to another PR.

The reason why the `limits_forcedirs` is two vectors is to make the prismatic joint reduction of DoFs part of the constraint Jacobian so that the limit constraint does not increase the error of the other constraints whenever there is numerical drift making the axes not aligned. Though it's a bit over-simplified right now: the correct modeling is probably to project `axis1` on the kernel of the basis of the locked DoFs of the second prismatic joint to get something like this: `let axis2 = -(limit_forcedir1 - basis2 * basis2.transpose() * limit_forcedir1)`.

emilk: prismatic narrow fixed

src/dynamics/solver/joint_constraint/revolute_velocity_constraint.rs
emilk: ...or rather, I should use `basis1/basis2` here to get the rotation error to be two dimensional... somehow.

emilk: narrow version fixes

src/dynamics/solver/joint_constraint/prismatic_velocity_constraint.rs
```diff
@@ -179,35 +210,48 @@ impl PrismaticVelocityConstraint {
         /*
          * Setup limit constraint.
          */
-        let mut limits_forcedirs = None;
+        let limits_forcedir2 = axis2.into_inner(); // hopefully axis1 is colinear with axis2
```
emilk, line 213: The old code used two different axis here, which means the applied impulses were not opposite and equal, thus violating conservation of momentum. When the constraint is closed to satisfied `axis1≈axis2`, and the choice here is arbitrary. If the angle constraint is very violated, then there is no single vector that makes sense here, but one could consider taking the mean of axis1 and axis2 here, just so as not to favor one constraint over the other. For now I think this is an improvement over what already existed, since at least we conserve momentum.

src/dynamics/solver/joint_constraint/prismatic_velocity_constraint.rs
```diff
@@ -179,35 +210,48 @@ impl PrismaticVelocityConstraint {
...
             let dist = danchor.dot(&axis1);
 
-            // TODO: we should allow both limits to be active at
-            // the same time, and allow predictive constraint activation.
-            if dist < joint.limits[0] {
-                limits_forcedirs = Some((-axis1.into_inner(), axis2.into_inner()));
-                limits_rhs = anchor_linvel2.dot(&axis2) - anchor_linvel1.dot(&axis1);
-                limits_impulse = joint.limits_impulse;
-            } else if dist > joint.limits[1] {
-                limits_forcedirs = Some((axis1.into_inner(), -axis2.into_inner()));
```
emilk, line 198: this was the momentum-violating bug mentioned in the PR description

src/dynamics/solver/joint_constraint/ball_velocity_constraint.rs
```diff
@@ -40,17 +40,20 @@ impl BallVelocityConstraint {
         rb2: &RigidBody,
         joint: &BallJoint,
     ) -> Self {
-        let anchor1 = rb1.position * joint.local_anchor1 - rb1.world_com;
-        let anchor2 = rb2.position * joint.local_anchor2 - rb2.world_com;
+        let anchor_world1 = rb1.position * joint.local_anchor1;
+        let anchor_world2 = rb2.position * joint.local_anchor2;
+        let anchor1 = anchor_world1 - rb1.world_com;
+        let anchor2 = anchor_world2 - rb2.world_com;
```
emilk, line 46: The names here are wildly inconsistent between files. It seems more common that `let anchor1 = position1 * self.local_anchor1;` (e.g. in `ball_position_constraint.rs`). Also sometimes the words `local` and `world` are part of the names (yay!) but sometimes not (😢 ). My approach is to try to make the smallest change possible to minimize conflicts.

emilk: It seems what is here called `anchor1` is often called `r1`.

src/dynamics/solver/joint_constraint/revolute_velocity_constraint_wide.rs
```diff
@@ -107,9 +107,37 @@ impl WRevoluteVelocityConstraint {
...
+        ) * SimdReal::splat(params.velocity_solve_fraction);
+
+        let velocity_based_erp_inv_dt = params.velocity_based_erp_inv_dt();
+        if velocity_based_erp_inv_dt != 0.0 {
+            let velocity_based_erp_inv_dt = SimdReal::splat(velocity_based_erp_inv_dt);
+
+            let lin_err = anchor2 - anchor1;
+
+            let local_axis1 =
+                Unit::<Vector<_>>::from(array![|ii| joints[ii].local_axis1; SIMD_WIDTH]);
```
emilk, line 128: maybe a `UnitVector` type alias could be useful (similar as for `UnitQuaternion`)

src/dynamics/solver/joint_constraint/revolute_velocity_constraint.rs
```diff
@@ -90,9 +89,31 @@ impl RevoluteVelocityConstraint {
...
+
+        let velocity_based_erp_inv_dt = params.velocity_based_erp_inv_dt();
+        if velocity_based_erp_inv_dt != 0.0 {
+            let lin_err = anchor2 - anchor1;
+
+            let axis1 = rb1.position * joint.local_axis1;
+            let axis2 = rb2.position * joint.local_axis2;
+
+            let axis_error = axis1.cross(&axis2);
+            let ang_err = basis2.tr_mul(&axis_error) - basis1.tr_mul(&axis_error);
```
emilk, line 112: This doesn't quite work for some reason :(

emilk:
Ok, I think I finally got all the kinks worked out now

src/dynamics/solver/joint_constraint/revolute_velocity_constraint_wide.rs
sebcrozet: I agree, I will create an issue to add this to `nalgebra`.

sebcrozet (approved):
Thank you for this very complete PR! Everything looks OK.

> How should we handle allowed_linear_error?

I dragged that parameter from what I did in nphysics years ago (which used GJK/EPA for almost all its collision-detection), so it's been a while since I revisited its usefulness and its default value.

The main reason for this parameter is to avoid the situation where the distance between the two objects is exactly (or close zero). This zero-distance configuration can cause numerical issues in the GJK/EPA collision-detection algorithms which reduces stability when you have a large pile of objects.

Let's not do anything about it in this PR as it it out of its scope. but here are various ways to handle this:
- See if a smaller value continues to work properly for, e.g, piles of convex polyhedron.
- Enable this threshold only for contacts involving problematic pairs of shapes (cones, cylinders, rounded shapes, and convex polyhedron).
- The best way to handle this would be to get rid of GJK/EPA to use something more robust. I have a couple of ideas of what we could use instead, but we are not there yet.

> ...though, is there even a test barrage? There is src_testbed, but I haven't found any docs or explanation for it. Seems to be only for manual testing?

Right now there is only manual testing by running for example `cargo run --release --bin all_examples3`.

PR #118, opened 2021-02-17
Add ColliderHandle:s to PairFilterContext

emilk:
This makes it easy to write custom filters based on the handles.

For instance, a user can add pairs of `ColliderHandle`s into a `HashSet` (or `Coarena`) and do a lookup in a `ContactPairFilter` to easily (if maybe not efficiently) disable collisions between pairs of colliders.

The same could be accomplished by using the `user_data: u128` field of the `Collider`, but it would be less ergonomic, and also with this PR that `user_data` space can be saved for more important things.


sebcrozet:
Thanks!

PR #119, opened 2021-02-22
Add joint motors

### LINKED ISSUE #87: Feature request: joint drives
Joint drives (or joint motors) allows one to simulate e.g. a motor turning a wheel, a muscle bending an arm, or a pushing piston.

In particular, I'd like to be able to set up angular forces for ball joints and revolute joints in an "angular spring" fashion, i.e.:

```
torque = stiffness * (target_angle - relative_angle)
       + damping * (target_angular_velocity - relative_angular_velocity)
```

(the word `damping` really only makes sense for when `target_angular_velocity=0` - some more clever nomenclature may be warranted).

In addition, it would be desireable to clamp this to a given maximum absolute torque.

For a revolve joint this would then be five parameters (`target_angle, target_angular_velocity, stiffness, damping, max_torque`). One could then use `target_angle` and `stiffness` to move the hands of a clock, and `target_velocity` + `damping` to turn the wheels of a car.

For ball joints we have three DOFs, so `target_angle` would need to be a quaternion and `target_angular_velocity` a vector. For the `siffness`, `damping` and `max_torque` I think one dimension is enough for most use cases (i.e. use the same strength for all three DOFs).

Related to this is a generalization of all joints as one 6DOF joint. If one can set a target angle and velocity as well as stiffness and damping for all the six degrees of freedom independently, then one actually has a general joint constraint. Basically one would lock a degree of freedom by setting a very high stiffness and damping to one dimension (and zero target velocity). This may simplify the inner workings of rapier in the long run, but probably be more work in the short run.

---

Of course, all this can be approximated using `RigidBody::apply_torque` and friends, but it really ought to be part of the constraint solver, as forces and torques tend to "lose" against real constraints (after all, one can approximate all the constraints using forces and torques).

sebcrozet:
This adds motors to the prismatic, revolute, and ball joints. The motor acts as a spring-like constraint, configurable as described in #87. As such, it is possible to give the motor a target velocity as well as a target position.

Multiple spring-like equations (velocity-based, acceleration-based, or force-based) can be chosen by the user in a per-joint basis.

The `joint3` demo has been updated to add motorized joint examples.

This PR also includes some preliminary work on a generic 6DoF joint as suggest in #87. However it's not implemented in a very satisfying way yet so its sources are commented out right now.

Fix #87

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,10 +18,11 @@ members = [ "build/rapier2d", "build/rapier2d-f64", "build/rapier_testbed2d", "e
 #nalgebra = { path = "../nalgebra" }
 
 #kiss3d = { git = "https://github.com/sebcrozet/kiss3d" }
-#parry2d = { git = "https://github.com/sebcrozet/parry" }
-#parry3d = { git = "https://github.com/sebcrozet/parry" }
-#parry2d-f64 = { git = "https://github.com/sebcrozet/parry" }
-#parry3d-f64 = { git = "https://github.com/sebcrozet/parry" }
+nalgebra = { git = "https://github.com/dimforge/nalgebra", branch = "dev" }
+parry2d = { git = "https://github.com/dimforge/parry" }
+parry3d = { git = "https://github.com/dimforge/parry" }
+parry2d-f64 = { git = "https://github.com/dimforge/parry" }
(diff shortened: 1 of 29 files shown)


extrawurst:
this is so cool! can't wait to put my hands on this!

Issue #121, opened 2021-02-24
Modify existing joints


emilk:
Now that joint motos are merged (https://github.com/dimforge/rapier/pull/119) it would be beneficial to be able to modify existing joints, e.g. to change the joint motor targets. Unlike `RigidBodySet` and `ColliderSet`, `JointSet` has no `get_mut` member. I don't know if this is an oversight, or a deliberate choice.

sebcrozet:
Yes, this is an oversight. There is a `JointSet::iter_mut` method but we are missing `JointSet::get_mut` and `JointSet::get_unknown_gen_mut`.

PR #123, opened 2021-02-24
Add JointSet::get_mut and get_unknown_gen_mut

emilk:
Fix https://github.com/dimforge/rapier/issues/121

### DIFF
--- a/src/dynamics/joint/joint_set.rs
+++ b/src/dynamics/joint/joint_set.rs
@@ -76,6 +76,12 @@ impl JointSet {
         self.joint_graph.graph.edge_weight(*id)
     }
 
+    /// Gets a mutable reference to the joint with the given handle.
+    pub fn get_mut(&mut self, handle: JointHandle) -> Option<&mut Joint> {
+        let id = self.joint_ids.get_mut(handle.0)?;
+        self.joint_graph.graph.edge_weight_mut(*id)
+    }
+
     /// Gets the joint with the given handle without a known generation.
     ///
     /// This is useful when you know you want the joint at position `i` but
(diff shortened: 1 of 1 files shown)


src/dynamics/joint/joint_set.rs
```diff
@@ -76,6 +76,12 @@ impl JointSet {
         self.joint_graph.graph.edge_weight(*id)
     }
 
+    /// Gets a mutable reference to the joint with the given handle.
+    pub fn get_mut(&mut self, handle: JointHandle) -> Option<&mut Joint> {
+        let id = self.joint_ids.get_mut(handle.0)?;
```
sebcrozet, line 81: This does not have to be mutable:

```suggestion
        let id = self.joint_ids.get(handle.0)?;
```

src/dynamics/joint/joint_set.rs
```diff
@@ -93,6 +99,23 @@ impl JointSet {
...
+    ///
+    /// This is useful when you know you want the joint at position `i` but
+    /// don't know what is its current generation number. Generation numbers are
+    /// used to protect from the ABA problem because the joint position `i`
+    /// are recycled between two insertion and a removal.
+    ///
+    /// Using this is discouraged in favor of `self.get_mut(handle)` which does not
+    /// suffer form the ABA problem.
+    pub fn get_unknown_gen_mut(&mut self, i: usize) -> Option<(&mut Joint, JointHandle)> {
+        let (id, handle) = self.joint_ids.get_unknown_gen_mut(i)?;
```
sebcrozet, line 112: This does not have to be mutable:

```suggestion
        let (id, handle) = self.joint_ids.get_unknown_gen(i)?;
```

sebcrozet (changes requested):
Thank you for this PR!

src/dynamics/joint/joint_set.rs
emilk: ah no, of course

sebcrozet (approved):

sebcrozet:
Thanks!

Issue #125, opened 2021-02-25
Joint motor stability


emilk:
I've been trying out [the new joint motors](https://github.com/dimforge/rapier/pull/119) (thanks for adding those!).

I've found a stability issue, which should be easy to resolve, but I wanted to make an issue before I jump to PR.

The issue is that for high values of stiffness and damping the motor will overshoot its target, producing instabilities. Here is an excerpt from <https://github.com/dimforge/rapier/blob/master/src/dynamics/solver/joint_constraint/revolute_velocity_constraint.rs#L113:L121>:

``` rust
        if stiffness != 0.0 {
            motor_angle = joint.estimate_motor_angle(&rb1.position, &rb2.position);
            motor_rhs += (motor_angle - joint.motor_target_pos) * stiffness;
        }

        if damping != 0.0 {
            let curr_vel = rb2.angvel.dot(&motor_axis2) - rb1.angvel.dot(&motor_axis1);
            motor_rhs += (curr_vel - joint.motor_target_vel) * damping;
        }
```

I think the `stiffness` here (which comes from `combine_coefficients`) should be clamped to the `[0, 1/dt]` interval, and `damping` should be clamped to `[0, 1]` (or possibly to `[0, velocity_based_erp_inv_dt]` and `[0, velocity_solve_fraction]` respectively once https://github.com/dimforge/rapier/pull/116 is merged). This should prevent overshoot while allowing the motor constraint to satisfy as quickly as possible. With such clamping the user can specify an infinite stiffness and damping and have it mean "as quickly as any other constraint"

sebcrozet:
Hi! As you suggest in #126, I feel like we should not clamp the value and simply document what the recommended values are. This would avoid some needless additional complication in our code, and this clamping could prevent some use-cases where, for example, one would want to setup an overly stiff spring to demonstrate how it behaves (for educational purposes).

sebcrozet:
The modeling of joints in Rapier has changed dramatically since this issue was open, and the suggested clamping is no longer relevant. I think we can close this and open a different issue if problems are found with the new implementation.

PR #126, opened 2021-02-26
Clamp motor strength to prevent instabilities

emilk:
Fix https://github.com/dimforge/rapier/issues/125

An alternative is to leave the clamping to the user, but then the safe ranges for stiffness and damping should be clearly documented for each `SpringModel` type. Maybe that's a better idea. 🤷 

I'm not quite sure what is the right thing to do here for `SpringModel::ForceBased` since I'm a bit unsure how it works, or should work. Seems to me `ForceBased` should mean that e.g. two revolute joints with the same motor settings attached to two bodies of different inertia's should lead to different angular accelerations, but afaict there is nothing taking inertia into account for `ForceBased` and in a test I wrote I couldn't get `SpringModel::ForceBased` to work as expected. (personally I think I will stick to using `VelocityBased`)

### DIFF
--- a/src/dynamics/joint/spring_model.rs
+++ b/src/dynamics/joint/spring_model.rs
@@ -1,4 +1,4 @@
-use crate::math::Real;
+use crate::{dynamics::IntegrationParameters, math::Real};
 
 /// The spring-like model used for constraints resolution.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


emilk:
As discussed in https://github.com/dimforge/rapier/issues/125#issuecomment-788922367 we should add some documentation instead

Issue #77, opened 2020-12-13
Far away object causes panic in broad phase


e00E:
> thread 'main' panicked at 'assertion failed: proxy.aabb.maxs[dim] >= self.min_bound', .cargo/registry/src/github.com-1ecc6299db9ec823/rapier3d-0.4.2/src/geometry/broad_phase_multi_sap.rs:172:13
stack backtrace:
   0: std::panicking::begin_panic
   1: rapier3d::geometry::broad_phase_multi_sap::SAPAxis::batch_insert
   2: rapier3d::geometry::broad_phase_multi_sap::BroadPhase::update_regions
   3: rapier3d::geometry::broad_phase_multi_sap::BroadPhase::find_pairs
   4: rapier3d::pipeline::physics_pipeline::PhysicsPipeline::step
   5: bug_tests::main

This can happen when an object is free falling without a floor.

If the panic is intentional it should be documented under what conditions it can occur and what users can do to prevent it.

slasktotten:
We started running into this as well and agree. Some more information regarding what might be causing it would be super helpful.

aristaeus:
I've written an example to reproduce this issue.

```rust
use rapier3d_f64::dynamics::*;
use rapier3d_f64::geometry::*;
use rapier3d_f64::pipeline::*;
use rapier3d_f64::math::*;

pub struct DummyEventHandler;
impl EventHandler for DummyEventHandler {
    fn handle_intersection_event(&self, _: IntersectionEvent) {
    }

    fn handle_contact_event(&self, _: ContactEvent) {
    }
}

fn main() {
    let mut bodies = RigidBodySet::new();
    let mut colliders = ColliderSet::new();
    let mut joints = JointSet::new();
    let mut pipeline = PhysicsPipeline::new();
    let mut broad_phase = BroadPhase::new();
    let mut narrow_phase = NarrowPhase::new();
    let integration_params = IntegrationParameters::default();

    let rigid_body = RigidBodyBuilder::new_dynamic().position(Isometry::translation(0.0, 5.0e10, 0.0));
    let rigid_body = bodies.insert(rigid_body.build());
    colliders.insert(ColliderBuilder::ball(1.0).build(), rigid_body, &mut bodies);

    pipeline.step(
        &Vector::zeros(),
        &integration_params,
        &mut broad_phase,
        &mut narrow_phase,
        &mut bodies,
        &mut colliders,
        &mut joints,
        None,
        None,
        &DummyEventHandler
    );
}
```

[...]

PR #128, opened 2021-03-01
Improve distant object panics in broad phase

aristaeus:
This improves the situation in #77:

* The region key has been replaced by an `i64` in the `f64` version of the library, increasing the range before panics occur.
* Asserts have been added to panic when creating the region key, with a comment explaining the issue to users still affected.


Vrixyz:
I updated the branch here: https://github.com/Vrixyz/rapier/tree/broad_phase_panic_fix-updated

Vrixyz:
Thanks!

PR #135, opened 2021-03-02
Remove unused dependency

emilk:
There may be others, but my go-to tool for this `cargo +nightly udeps` doesn't work at the moment due to the current build-issue.


sebcrozet:
Thank you for this PR! I didn't know about `udeps`, this is very nice.
The CI is failing right now because of an issue with the compilation of the testbed after the upgrade to nalgebra 0.6.1. It will be fixed as part of #136.

sebcrozet:
Thanks!

Issue #132, opened 2021-03-02
Per-collider mass-properties


emilk:
Currently the mass properties of a collider is computed from its shape and density. For some shapes, such as `TriMesh`, this is always zero.

I'd like to be able to (optionally) specify the `MassProperties` per `ColliderBuilder`. This will make it easy to override or specify it per collider, and still have rapier compute the final `MassProperties` of the `RigidBody` automatically.

# Proposal
**EDIT**: the source is easier to read: https://github.com/dimforge/rapier/pull/133

I propose we add `mass_properties: MassProperties` member to `Collider`. It (like all other members) will be set at initialization and never changed.

To `ColliderBuilder` we add a `mass_properties: Option<MassProperties>` member and a function to set it. In `ColliderBuilder::build` we check if `mass_properties` is set, and if so assign it to `Collider::mass_properties` (and also use it to compute the average `density` so we don't have to remove the `Collider::density()` member). If `ColliderBuilder::mass_properties` is not set, we compute the `mass_properties` from `Shape::mass_properties` and `ColliderBuilder::get_density`

sebcrozet:
Hi! Thank you for this proposal. This is a great idea from a user-experience point of view, but I'm not sure this is the best approach considering the potential memory overhead this implies. In particular, I feel like setting the mass properties explicitly is something most users won't do, which makes the memory overhead less justifiable.

Right now, you can use `RigidBodyBuilder::mass_properties` to set the initial mass properties of a rigid-body (I say "initial" because the collider's mass properties will be added to that afterwards). So the typical way of having full control over the mass-properties of a rigid-body right now is to set it with `RigidBodyBuilder::mass_properties` and use 0 for the density of the colliders attached to it.

Here are some though on how we could avoid this memory overhead:
- Instead of modifying colliders, we could just make it easier to compose two `MassProperties`, and add a new `MassProperties` contribution to an existing `RigidBody`. That way the user just has to call ` shape.mass_properties()` to get the properties of a shape, and use the improve rigid-body API to take that contribution into account.
- Or we could modify the `ColliderBuilder` only, so it contains a `MassProperties` field. And that field would be taken into account by `ColliderSet::insert` to update the mass-properties of its parent rigid-body. But the collider itself never contains that `MassProperties` field.
- Or we could add an `Option<Box<MassProperties>>` field to colliders. That way it won't consume any significant memory in the most common case where the user just relies on the collider density instead of setting the mass-properties explicitly.

[...]

PR #139, opened 2021-03-04
Call update_world_mass_properties from RigidBody::set_mass_properties

emilk:
I think this was a bug. Related to https://github.com/dimforge/rapier/issues/132


sebcrozet:
Thanks!

PR #140, opened 2021-03-05
Derive Clone for JointSet

Waridley:
`JointSet` was inadvertently left out of https://github.com/dimforge/rapier/pull/54 when `Clone` was implemented for most other types.


sebcrozet:
Thanks!

PR #141, opened 2021-03-05
Per collider mass props

emilk:
Fix https://github.com/dimforge/rapier/issues/132

This uses the `Option<Box<MassProperties>>` idea to save memory in the common case.

### DIFF
--- a/src/geometry/collider.rs
+++ b/src/geometry/collider.rs
@@ -49,6 +49,7 @@ impl ColliderFlags {
 pub struct Collider {
     shape: SharedShape,
     density: Real,
+    mass_properties: MassProperties,
     pub(crate) flags: ColliderFlags,
     pub(crate) solver_flags: SolverFlags,
     pub(crate) parent: RigidBodyHandle,
@@ -134,9 +135,9 @@ impl Collider {
     //     aabb1.merged(&aabb2)
     // }
 
-    /// Compute the local-space mass properties of this collider.
-    pub fn mass_properties(&self) -> MassProperties {
-        self.shape.mass_properties(self.density)
+    /// Read the local-space mass properties of this collider.
+    pub fn mass_properties(&self) -> &MassProperties {
+        &self.mass_properties
     }
 }
 
@@ -146,8 +147,11 @@ impl Collider {
 pub struct ColliderBuilder {
     /// The shape of the collider to be built.
     pub shape: SharedShape,
-    /// The density of the collider to be built.
+    /// The uniform density of the collider to be built.
     density: Option<Real>,
+    /// Overrides automatic computation of `MassProperties`.
+    /// If None, it will be computed based on shape and desnity.
+    mass_properties: Option<MassProperties>,
     /// The friction coefficient of the collider to be built.
     pub friction: Real,
     /// The rule used to combine two friction coefficients.
@@ -177,6 +181,7 @@ impl ColliderBuilder {
         Self {
             shape,
             density: None,
+            mass_properties: None,
             friction: Self::default_friction(),
             restitution: 0.0,
             delta: Isometry::identity(),
@@ -456,6 +461,8 @@ impl ColliderBuilder {
     }
 
     /// Sets whether or not the collider built by this builder is a sensor.
+    /// Sensors will have a default density of zero,
+    /// but if you call [`Self::mass_properties`] you can assigna a mass to a sensor.
     pub fn sensor(mut self, is_sensor: bool) -> Self {
         self.is_sensor = is_sensor;
         self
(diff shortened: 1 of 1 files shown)


src/geometry/collider.rs
```diff
@@ -49,6 +49,8 @@ impl ColliderFlags {
 pub struct Collider {
     shape: SharedShape,
     density: Real,
+    /// If None, use [`Self::density`] and [`SharedShape::mass_properties`].
+    mass_properties: Option<Box<MassProperties>>,
```
sebcrozet, line 53: Since both `density` and `mass_properties` are exclusive, let's make this explicit by using [either](https://crates.io/crates/either): `density_or_mass_properties: Either<Real, Box<MassProperties>>`

That way we are sure that we won't forget to use the correct one, and we don't have to do all the `mass(&mp) / volume` computation added by this PR.

This will require one breaking change: `Collider::density()` will have to return an `Option<Real>`, indicating that its `MassProperties` have been set explicitly.

src/geometry/collider.rs
```diff
@@ -146,8 +151,11 @@ impl Collider {
 pub struct ColliderBuilder {
     /// The shape of the collider to be built.
     pub shape: SharedShape,
-    /// The density of the collider to be built.
+    /// The uniform density of the collider to be built.
     density: Option<Real>,
+    /// Overrides automatic computation of `MassProperties`.
+    /// If None, it will be computed based on shape and desnity.
```
sebcrozet, line 157:
```suggestion
    /// If None, it will be computed based on shape and density.
```

src/geometry/collider.rs
```diff
@@ -456,6 +465,8 @@ impl ColliderBuilder {
     }
 
     /// Sets whether or not the collider built by this builder is a sensor.
+    /// Sensors will have a default density of zero,
+    /// but if you call [`Self::mass_properties`] you can assigna a mass to a sensor.
```
sebcrozet, line 469:
```suggestion
    ///
    /// Sensors will have a default density of zero,
    /// but if you call [`Self::mass_properties`] you can assign a mass to a sensor.
```

src/geometry/collider.rs
```diff
@@ -492,12 +503,22 @@ impl ColliderBuilder {
         self
     }
 
-    /// Sets the density of the collider this builder will build.
+    /// Sets the uniform density of the collider this builder will build.
+    /// This will be overridden by a call to [`Self::mass_properties`] so it only makes sense to call
+    /// either [`Self::density`] or [`Self::mass_properties`].
```
sebcrozet, line 508:
```suggestion
    /// Sets the uniform density of the collider this builder will build.
    ///
    /// This will be overridden by a call to [`Self::mass_properties`] so it only makes sense to call
    /// either [`Self::density`] or [`Self::mass_properties`].
```

src/geometry/collider.rs
```diff
@@ -492,12 +503,22 @@ impl ColliderBuilder {
...
     pub fn density(mut self, density: Real) -> Self {
         self.density = Some(density);
         self
     }
 
+    /// Sets the mass properties of the collider this builder will build.
+    /// If this is set, [`Self::density`] will be ignored, so it only makes sense to call
+    /// either [`Self::density`] or [`Self::mass_properties`].
```
sebcrozet, line 516:
```suggestion
    /// Sets the mass properties of the collider this builder will build.
    ///
    /// If this is set, [`Self::density`] will be ignored, so it only makes sense to call
    /// either [`Self::density`] or [`Self::mass_properties`].
```

src/geometry/collider.rs
```diff
@@ -540,7 +561,21 @@ impl ColliderBuilder {
...
+        if let Some(mp) = self.mass_properties {
+            mass_properties = Some(Box::new(mp));
+
+            let volume = volume(&self.shape);
+            density = if volume == 0.0 || mp.inv_mass == 0.0 {
+                Real::INFINITY
+            } else {
+                mass(&mp) / volume
+            };
+        } else {
```
sebcrozet, line 574: Switching to `Either` will make this unnecessary.

src/geometry/collider.rs
```diff
@@ -570,3 +606,11 @@ impl ColliderBuilder {
...
     }
 }
+
+fn volume(shape: &SharedShape) -> Real {
+    mass(&shape.mass_properties(1.0)) // TODO: add SharedShape::volume to parry
+}
+
+fn mass(mp: &MassProperties) -> Real {
+    crate::utils::inv(mp.inv_mass) // TODO: add MassProperties::mass() to parry
+}
```
sebcrozet, line 616: Switching to `Either` will make this unnecessary. Though I agree it would be nice to add to Parry what the `TODO`s are suggesting.

sebcrozet (changes requested):
Thank you for this PR!
Re-parenting a collider is indeed something worth handling correctly.

sebcrozet (approved):

sebcrozet:
All good, thanks!

PR #144, opened 2021-03-06
Fix failure in collision pipeline stepping due to hardcoded island size

### LINKED ISSUE #143: Stepping a collision pipe always panics due to island size being <1
## Problem
Stepping a collision pipeline always results in the following error:

`
thread 'main' panicked at 'The minimum island size must be at least 1.', /Users/rishflab/rapier/build/rapier2d/../../src/dynamics/rigid_body_set.rs:508:9
`

How I created the collision pipeline:

```rust
let mut pipeline = CollisionPipeline::new();
let mut broad_phase = BroadPhase::new();
let mut narrow_phase = NarrowPhase::new();
let mut bodies = RigidBodySet::new();
let mut colliders = ColliderSet::new();
let (contact_send, contact_recv) = channel::unbounded();
let (intersection_send, intersection_recv) = channel::unbounded();
let event_handler = ChannelEventCollector::new(intersection_send, contact_send);

pipeline.step( 0.1, &mut broad_phase, &mut narrow_phase, &mut bodies, &mut colliders,&(), &event_handler);
```

## Diagnosis
This is because the minimum island size is hardcoded to 0 and there is an assertion that the min island size should be >0

## Proposed solution
Pass the island size in as a parameter to `step()`

rishflab:
The min island size was hardcoded as 0 which was causing the assertion
that the min_island_size > 0 to always fail. Added a min island size
param to the step function to prevent this error.

Closes #143

### DIFF
--- a/src/pipeline/collision_pipeline.rs
+++ b/src/pipeline/collision_pipeline.rs
@@ -43,6 +43,7 @@ impl CollisionPipeline {
         colliders: &mut ColliderSet,
         hooks: &dyn PhysicsHooks,
         events: &dyn EventHandler,
+        min_island_size: usize,
     ) {
         bodies.maintain(colliders);
         self.broadphase_collider_pairs.clear();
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thank you for this fix. I made a slightly different fix in #157 that does not expose the `min_island_size` parameter to the end-user. So this PR can be closed now.

PR #149, opened 2021-03-08
Fix regression in the WASM version of the velocity solver.

sebcrozet:


emilk:
I tested this, and it helps, but there seems to be more problems lingering

sebcrozet:
> I tested this, and it helps, but there seems to be more problems lingering

Can you perhaps be more specific? With a gif of the problem perhaps?

emilk:
Sorry, my mistake. There is a problem with the collisions between capsules and trimeshes but that is existing in older rapier too, so it is a separate problem.

PR #153, opened 2021-03-13
Fix scalability problem in the broad-phase when big colliders are involved

sebcrozet:
The Multi-SAP method we currently use for the broad-phase has one significant scalability problem: large objects intersecting lots of regions, result in poor performances and very high memory usage.

So a solution implemented in this PR is to combine SAP with a hierarchical grid. A hierarchical grid is composed of several layers. And each layer have different region sizes. For example all the regions on layer 0 will have the size 1x1x1. All the regions on the layer 1 will have the size 10x10x10, etc. That way, a given AABB will be inserted on the layer that has regions big enough to avoid the large-object problem. For example a 20x20x20 object will be inserted in the layer with region of size 10x10x10, resulting in only 8 regions being intersect by the AABB. (If it was inserted in the layer with regions of size 1x1x1, it would have intersected 8000 regions, which is a problem performancewise.)


sebcrozet:
This PR is now closed because it was part of #157 which is merged now.

PR #154, opened 2021-03-19
Provide access to rigid body flags

### LINKED ISSUE #150: Add a way to get rigid body flags
Currently there is no way to extract rigid body flags, this makes impossible correct serialization of rigid bodies using custom serializers.

mrDIMAS:
Fixes #150


sebcrozet:
Thank you for this PR! We should not expose the flags themselves to the user.
Instead we should provide more descriptive methods: `fn is_translation_locked() -> bool`, `is_rotation_locked() -> [bool; 3]`.

sebcrozet:
It appears that I don't have the rights to modify your branch. Since I intend to add this to this week's release, I took the liberty to open another PR that replaces this one: #160

PR #173, opened 2021-04-12
Add #[must_use] to builders, expose more fields

DasEtwas:
I nearly forgot that `RigidBodyBuilder` and `ColliderBuilder` use `mut self` semantics in their parameters setting functions, so a `#[must_use]` seems appropriate.

Also, `RigidBodyBuilder` seemed surprisingly locked-down in terms of field visibilities compared to `ColliderBuilder`. `body_status` is set upfront from the builder's new_* constructors, and `flags` appears to be a crate-private type, so everything else was made public by this PR.

`RigidBodyBuilder` seems like it contains pretty plain data, which permits deriving Copy. Debug should be fine as well as the flags are nicely formatted by `bitflags`.

For my use-case, the builders also serve as creation templates, so having (read-)access to everything is nicer than recreation.

Please let me know if making these fields private was a conscious design decision.

### DIFF
--- a/src/dynamics/rigid_body.rs
+++ b/src/dynamics/rigid_body.rs
@@ -564,7 +564,7 @@ impl RigidBody {
 }
 
 /// A builder for rigid-bodies.
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 #[must_use = "Builder functions return the updated builder"]
 pub struct RigidBodyBuilder {
     /// The initial position of the rigid-body to be built.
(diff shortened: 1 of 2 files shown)


Waridley:
Isn't deriving `Copy` kind of contrary to move semantics and `must_use`? I am strongly in favor of deriving `Clone`, but `Copy` could result in surprising behavior if a builder gets moved into separate setter functions and the compiler just copies the data instead of mutating it in-place.

DasEtwas:
Yes, I didn't catch that.

DasEtwas:
All set.

sebcrozet:
Thanks!

PR #176, opened 2021-04-17
Created a method to retrieve the SharedShape from a collider

sdfgeoff:
In my case this is for creating a compound shape from a set of existing colliders.

I do find if confusing that the `Collider.shape()` method returns a `dyn Shape`, the `Collider.set_shape()` requires a `SharedShape`, but fixing this such that `shape()`  returns a SharedShape would be a breaking change. Perhaps adding a `Collider.set_shared_shape()` method and deprecating `Collider.set_shape()`?

### DIFF
--- a/src/geometry/collider.rs
+++ b/src/geometry/collider.rs
@@ -244,6 +244,11 @@ impl Collider {
         self.shape = shape;
     }
 
+    /// Retrieve the SharedShape. Also see the `shape()` function
+    pub fn shared_shape(&self) -> &SharedShape {
+        &self.shape
+    }
+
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks!
The mean reasons why `Collider::shape()` returns `&dyn Shape` is because `Collider::shape_mut()` returns a `&dyn mut Shape`. I'm not sure how confusing `Collider::set_shape` is considering we couldn't take a `&dyn Shape` anyway. I'll keep it as it is right now, and see if we get other reports of confusion on that matter.

PR #193, opened 2021-05-29
Testbed: check if the egui context wants pointer focus, disable orbit camera if so.

rezural:
This disables egui interactions being sent to the orbital camera if mouse events happen within egui.

i.e. scrolling on the egui doesnt affect the orbitalcamera's zoom.

This may also want to check for 

```
|| ui_context.ctx().wants_keyboard_input()

```

not sure about this.


sebcrozet:
Thanks!
Mmh, not sure about the keyboard inputs either. We'll see if we have any issue in practice.

PR #195, opened 2021-06-02
Add feature flag to disable floating point exceptions in unexceptional code

tpdickso:
This PR adds the feature flag `avoid-fe-exceptions` to rapier. When set, this flag will temporarily disable floating-point exception handling while executing code which would otherwise generate floating-point exceptions in its normal operation. This improves the debugging experience for the end-user, who can use `feenableexcept` to determine where their code may be incorrectly generating infinite or NaN values that can cause the simulation to panic, without the program terminating early due to a floating-point exception raised as part of normal operation.

Currently, the only places I've found where this occurs are when normalizing a vector which may have a norm of zero, and handling the resulting infinite/NaN values afterward. There may be other parts of the code where this should be expanded to; but since this is just to make debugging friendlier, that can probably just be dealt with through bug reports as they arise.

An example of how to use this is to put this code in your application binary during its `main` execution:
```
extern "C" {
    fn feenableexcept(flags: i32);
}
...
feenableexcept(9);
```
This code will cause rapier to crash during normal execution when the `avoid-fe-exceptions` flag isn't set. If the `avoid-fe-exceptions` flag is set, then this will (hopefully) only crash if something goes wrong in the simulation, such as the introduction of a NaN value, for example by applying an infinite force to an object with infinite mass.


src/utils.rs
```diff
@@ -670,6 +670,52 @@ impl Drop for FlushToZeroDenormalsAreZeroFlags {
...
+    #[cfg(not(feature = "avoid-fe-exceptions"))]
+    pub fn disable_floating_point_exceptions() -> Self {
+        Self { }
+    }
+
+    #[cfg(feature = "avoid-fe-exceptions")]
+    pub fn disable_floating_point_exceptions() -> Self {
+        unsafe {
+            let mut original_flags = [0; 256];
+            feholdexcept(original_flags.as_mut_ptr() as *mut _);
```
tpdickso, line 704: This particular function creates a copy of the floating-point environment in which the program is running and stores it in the destination pointer, and continues forward with all exceptions disabled. In theory, this could cause spooky errors if the floating-point environment is mutated while this object is alive, since it will be restored back to its previous state when this object is dropped. Ideally though I don't think that's a likely problem.

src/utils.rs
```diff
@@ -670,6 +670,52 @@ impl Drop for FlushToZeroDenormalsAreZeroFlags {
...
+    // (not even in the libc crate, as of the time of writing.)
+    // But since the state is intended to be stored on the stack,
+    // 256 bytes should be more than enough.
+    original_flags: [u8; 256],
+}
+
+#[cfg(feature = "avoid-fe-exceptions")]
+extern "C" {
+    fn feholdexcept(env: *mut std::ffi::c_void);
+    fn fesetenv(env: *const std::ffi::c_void);
```
tpdickso, line 691: Technically, these functions will only exist on Linux (maybe Mac? I don't know!) so only checking for the feature flag is insufficient here. However, the windows version of this code would be completely different, and I haven't really got a good windows development environment to figure that out on. However, thinking about it, I think the way it is now is actually good -- simply compiling it out if the flag is set on windows might silently confuse users. If the linker fails when the feature flag is enabled on windows, that's a good sign not to use it yet, because it isn't yet implemented on windows.

But maybe there's a better way to just disable the flag on windows. I'm new to feature flags!

src/utils.rs
```diff
@@ -670,6 +670,52 @@ impl Drop for FlushToZeroDenormalsAreZeroFlags {
...
+// code that is generating NaNs and infinite values erroneously.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct DisableFloatingPointExceptionsFlags {
+    #[cfg(feature = "avoid-fe-exceptions")]
+    // We can't get a precise size for this, because it's of type
+    // `fenv_t`, which is a definition that doesn't exist in rust
+    // (not even in the libc crate, as of the time of writing.)
+    // But since the state is intended to be stored on the stack,
+    // 256 bytes should be more than enough.
+    original_flags: [u8; 256],
```
tpdickso, line 685: Unfortunately, the code I'd hoped to use which simply checked the status of the exception flags doesn't seem to be possible; there isn't a function for just retrieving the state of the flags into a `c_int`. The only way to actually retrieve and save the state is to store it inside of a `fenv_t`, which is a structure definition which is platform-dependent and could theoretically have an arbitrary size. The functions `fegetexceptflag` and `fetestexcept` appear to refer to the *state* of the floating-point exceptions (i.e., whether they have been raised), and *not* whether or not those exceptions are *enabled*.

So, this code is definitely not ideal for it, since if you enable it on a platform where a `fenv_t` is larger than 256 bytes, it will just explode your stack. But since this feature flag is only meant for debugging I feel like that's okay for now, even if I hate it.

sebcrozet:
Thanks! The limitation sound acceptable. I'll merge this and fix the few outstanding issues that make the CI fail.
I'll also rename the `avoid-fe-exceptions` feature to`debug-disable-legitimate-fe-exceptions`.

sebcrozet:
(Sorry, I closed by mistake. It's merged now.)

PR #201, opened 2021-06-09
Derive Clone for IslandManager

Waridley:
`Clone` police here :laughing: Since `IslandManager` implements `Serialize`/`Deserialize`, I'm assuming it should also implement `Clone`. 

Also fixed a random typo I found.


sebcrozet:
Thanks!

PR #209, opened 2021-06-29
WIP: Enable plugin graphics

rezural:
This is the changes required to upgrade salva to rapier 0.9.

Graphics are not being rendered, this is just so that salva's FluidTestbedPlugin has access to the data it will need.

### DIFF
--- a/src_testbed/graphics.rs
+++ b/src_testbed/graphics.rs
@@ -195,9 +195,9 @@ impl GraphicsManager {
             .cloned()
             .unwrap_or_else(|| self.alloc_color(materials, handle, !body.is_dynamic()));
 
-        self.add_with_color(
+        let _ = self.add_with_color(
             commands, meshes, materials, components, handle, bodies, colliders, color,
-        )
+        );
     }
 
     pub fn add_with_color(
@@ -210,8 +210,7 @@ impl GraphicsManager {
         bodies: &RigidBodySet,
         colliders: &ColliderSet,
         color: Point3<f32>,
-    ) {
-        //        let body = bodies.get(handle).unwrap();
+    ) -> Vec<EntityWithGraphics> {
         let mut new_nodes = Vec::new();
 
         for collider_handle in bodies[handle].colliders() {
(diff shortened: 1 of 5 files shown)


src_testbed/graphics.rs
```diff
@@ -392,6 +393,10 @@ impl GraphicsManager {
     pub fn nodes_mut(&mut self) -> impl Iterator<Item = &mut EntityWithGraphics> {
         self.b2sn.values_mut().flat_map(|val| val.iter_mut())
     }
+
+    pub fn prefab_meshes(&self) -> &HashMap<ShapeType, Handle<Mesh>, RandomState> {
+        &self.prefab_meshes
+    }
 }
```
rezural, line 400: The FluidsTestbedPlugin will presumably need access to the prefab meshes (unless it wants to create its own ones, but this is DRY

src_testbed/testbed.rs
```diff
@@ -864,13 +867,15 @@ fn update_testbed(
...
-    mut gfx_components: Query<(&mut Transform,)>,
+    gfx_components: Query<(&mut Transform,)>,
     mut cameras: Query<(&Camera, &GlobalTransform, &mut OrbitCamera)>,
 ) {
     let meshes = &mut *meshes;
     let materials = &mut *materials;
     let prev_example = state.selected_example;
 
+    let gfx_components = Arc::new(Mutex::new(gfx_components));
+
```
rezural, line 878: I'm not sure if Arc<Mutex<>> is the way to go here, but passing the mutable transform is not possible without it. Not sure yet if the gfx_components will need to be a mutable query yet..

rezural (commented):
There's a couple of println's that snuck in here, I will remove them before this is ready

src_testbed/graphics.rs
rezural: I think this is no longer needed, I will remove it.

src_testbed/testbed.rs
sebcrozet: This is odd. I tried removing your `Arc<Mutex<>>` and it compiled fine. Did I miss something?

sebcrozet (changes requested):

rezural:
Probably just me fighting the borrow checker..?

rezural:
OK, I disabled it too, seems to work fine. Do you want me to push up the changes?

rezural (commented):
OK, I have updated this.

rezural:
Updated to make collider & rigid body optional

sebcrozet (approved):

sebcrozet:
Looks great, thanks!

PR #211, opened 2021-07-10
Derive and impl more traits for components

Waridley:
I mostly need `PartialEq` on a few `RigidBody` components, but I looked at all of the other component types and tried to determine whether they could implement `Eq`, `PartialOrd`, `Ord`, and/or `Hash`. Most of the types containing `Real`s only make sense to implement `PartialEq`, but some of the other components seem to be able to implement more traits. Also added `AsMut + DerefMut` to `ColliderPosition` for convenience since the field is public anyway, and found a few more typos.

Let me know if there's anything I missed or anything I added that I shouldn't have.


sebcrozet:
Thanks!

PR #213, opened 2021-07-11
Update intersection graph regardless of active_events

Ralith:
It looks like the contact graph updates unconditionally, whereas the intersection graph's edge labels only update if `ActiveEvents::INTERSECTION_EVENTS` is set. This is inconsistent and doesn't seem to be documented, so I'm guessing it wasn't intended.


sebcrozet (changes requested):
Hi! You are right, this wasn't intended. However, this changes will completely disable any events because the `intersection != edge.weight` condition will always fail now.

The `edge.weight = intersection` should happen after the `if` block instead.

Ralith:
Whoops, good catch! Fixed.

sebcrozet (approved):

PR #214, opened 2021-07-11
clear plugin graphics when examples changed, then clear the plugins vec

rezural:
Upstream salva upgrade changes need this to work when changing examples.


sebcrozet:
Thanks!

PR #220, opened 2021-07-26
Fix copy paste bug in ball joint docs

oli-obk:
The second anchor is expressed in the second body's local space


sebcrozet:
Thanks!

PR #221, opened 2021-07-26
Add a pointer from `docs.rs` docs to rapier docs

jpetkau:
Added a pointer from the module-level cargo docs to the actual docs on `rapier.rs` so other people won't be lost like I was.

(Searches for documentation often end up on `docs.rs`, which is not where most of Rapier's docs live. The README has a pointer to the docs, but it isn't as visible once you're on `docs.rs`.)

This is the same change as pr89 for `bevy_rapier`. This one is less important since it's harder to overlook `rapier.rs` from here, but I think it doesn't hurt.

Ideally you'd be able to just include the `README.md` into the lib documentation, but I think that's still a nightly-only feature in cargo doc.


sebcrozet:
Thanks!

PR #226, opened 2021-08-02
WIP: extract harness into rapier, so that graphics code is not compiled if it's not being used

rezural:
Edit: Clarification..

This extracts the harness from within the testbed, into rapier itself. It also extracts the PhysicsState struct, as that is needed by the harness.

Use Cases:
Good starting point for running a simulation in client code, without the testbed graphics.
Running a simulation on a game server.
Long running simulations, where I have found graphics driver errors sometimes arise after hours, probably due to rust<->graphics interactions, etc.
This also minimizes binary size when uploading to servers.

Changes;

cargo.toml changes for features

changes required to make having a harness_plugin & testbed_plugin ergonomic

documentation of ThreadState & RunState

extract PhysicsState and PhysicsEvents

change harness.state to harness.run_state


rezural:
I also havent checked the documentation output from this, to ascertain that it formats correctly etc.

I will do this soon.

sebcrozet:
Hi! Thank you for this PR. I don’t think that moving the harness into Rapier itself is a good thing. It adds complexity to the library with a somewhat little benefit for most users that will integrate Rapier as part of their own game/event loop.

I believe that I better approach would be to just move the graphics thing from the Rapier testbed behind a Cargo feature.

rezural:
Yeah, ok,

I did wonder if putting it in rapier proper was a good idea..

It seems complicated to hide the testbed graphics behind a feature, as it is all graphics. I havent had a look yet though, so I'll get back to you on this.

I'll probably just copy it externally, and maintain it separately without the graphics components. As I am currently the only user of the harness, this probably makes sense. It has been quite a work integrating it with the current testbed, when I am the only user of the functionality I extracted..

Thanks!

PR #234, opened 2021-08-28
Clarify coordinate space of TOI query results

Ralith:
These are confusing because the documentation of the `TOI` struct specifies local space, which doesn't seem to be the case here in practice, perhaps because the "local space" in question is considered to be the entire collision world acting as a single shape.


sebcrozet:
Thanks!

Issue #241, opened 2021-10-08
Feature Request: Spring


nascheinkman:
Hi,

I've been wanting to create an open source simulation of tensegrity structures and I've been wanting to use rapier to do it because its been really fun to use compared to other physics libraries I've tried. 

However, one of the things necessary for tensegrity simulation is some form of spring physics. I've noticed that rapier has a `SpringModel` but it seems like that's only used for joint motors. 

Is supporting springs something that the rapier team is interested in? I'm willing to write the code, but I would still need direction on how you would like it implemented, and how robust you would like it (perhaps like the `ForceGenerator` in nphysics to allow for nonlinear springs?). I already have a bare-bones working example trying to quickly hack it in as a joint, but after learning how joints are implemented in rapier it does not seem like it belongs there. 

I have already tried implementing a spring externally using `body.apply_force` after each physics step, but for some reason it added a bunch of energy to the system over time, and that issue did not happen when implemented the same way but internally as a joint.

sebcrozet:
The `ForceGenerator` in `nphysics` wasn’t complete enough to handle non-linear springs properly.
I believe it will be best to model a spring as a joint. My guess is that we could either:
- Create a whole new joint just for that.
- Or modify the `BallJoint` so it can work in "spring mode". This could, for example be achieved by doing the position correction at the velocity level instead of the position level (we already have code for the `BallJoint` to handle velocity-level position correction, but whether or not this is enabled depends on a global flag that affects the whole simulation.

nascheinkman:
I've added distance limits to the spring that seem to work pretty well and I fixed a few issues I found with my initial implementation. I still need to add the wide version, but I think the normal version is ready.

The draft pull request is here: https://github.com/dimforge/rapier/pull/255
I included an example for the spring joint that I've been using for testing and comparison.

nascheinkman:
The SpringJoint should be ready now here: https://github.com/dimforge/rapier/pull/255 . It includes limits and supports SIMD constraints.

chrislicodes:
Hey 👋 , when will this be merged in?

marwie:
Hello, wondering the same :)

Wolftousen:
@chrislicodes @marwie I wouldn't count on this getting merged at all.  The code was written before an overhaul of the engine happened and this code would need a lot of rework to get in now.

[...]

PR #255, opened 2021-10-26
Add Spring Joint

nascheinkman:
Adds a SpringJoint as discussed in https://github.com/dimforge/rapier/issues/241.

It also includes a `spring_joint` example that I've been using for testing. 

This is currently a work in progress. It still needs to implement a wide version and distance limits.

### DIFF
--- a/examples2d/all_examples2.rs
+++ b/examples2d/all_examples2.rs
@@ -23,6 +23,7 @@ mod polyline2;
 mod pyramid2;
 mod restitution2;
 mod sensor2;
+mod spring_joint2;
 mod trimesh2;
 
 fn demo_name_from_command_line() -> Option<String> {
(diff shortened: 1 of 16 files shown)


shenjackyuanjie:
how is this PR going?

nascheinkman:
> how is this PR going?

Sorry I forgot about this PR

As Wolftousen mentioned [here](https://github.com/dimforge/rapier/issues/241#issuecomment-1305005235), this PR was made before an overhaul of the engine, and I don't know how to implement springs in the new engine. 

I tried looking into attaching motors to Wolftousen's RopeJoints to see if I could create what would effectively be a SpringJoint, but I couldn't seem to get the motors to run. I'm not 100% sure but I think it's because of the TODO [here](https://github.com/dimforge/rapier/blob/958fba8ed43328e7e33bb9585a735577fa1061e1/src/dynamics/solver/joint_constraint/joint_velocity_constraint.rs#L199), and I don't know how to implement that TODO.

I'll close this for now.

PR #267, opened 2022-01-02
Implement multibody joints, and new velocity-based constraints solver

sebcrozet:
This is an absolutely massive PR containing about 4 months of work. The main focus of this work was to implement joints based on the "reduced-coordinates" modeling. These are joints that never violate their positional constraint.

When I started working on loop-closing joints (i.e. impulse joints attached to multibodies based on reduced-coordinates), it became quickly apparent that the current design and solver of Rapier would make this work an absolute nightmare: too many special-cases to handle, and the non-linearity of the position-based solver would make multibodies very inefficient.

## Multibody joints
I call `MultibodyJoint` joints that are based on the reduced-coordinates formalism. I worked hard on making the API of multibody joints as close as I can to the API of impulse-based joints (the ones we had so far), despite the fact that their underlying mathematical modeling are completely opposite. That way, it will be easy to switch and experiment between both representations, to pick the one that works best on a given setting.

The implementation of these joints isn’t quite finished yet (anything else than fixed/revolute/prismatic/spherical joints won’t work properly) but finishing the work won’t be very difficult (I already written down the math and just need to implement it now). So I will make an _alpha_  release of Rapier now, and add these fixes after (next week probably).

## Solver changes
I changed the constraints solver to only contain two linear velocity-based phases (instead of one velocity-based phase followed by a position-based phase):
1. **A biased velocity constraint resolution**, which solves for the contact and joint constraints and tries to correct
  penetrations and joint drifts at the same time. This will potentially add artificial energy to the mechanical system,
  resulting in potential jitters, and "popping" effects due to penetration.
2. **A non-biased _stabilization_ velocity resolution**, which removes most of the excessive energy introduced into the
   mechanical system.

With both phases combined, typical simulations won’t show any large "popping" effect or jitter, while feeling stiffer
overall. In addition, this solver doesn’t require any warmstarting, meaning that the force computation no longer
rely on the complete history of the simulation to converge.

## Joint changes
All the (impulse-based) joints have been replaced by a single generic joint `ImpulseJoint` structure commonly called "6-dofs joints" in other 3D physics engines. For reduced-coordinates joints, we have a similar `MultibodyJoint`. Each degree of freedom of the 6-dofs joint can be fixed/limited/motorized individually . Though motorization of two or three angular degrees of freedom is a bit fragile right now (some special cases have to be added to handle some numerical limitations).

In order to simplify the creation of an `ImpulseJoint` or `MultibodyJoint` for the usual cases, we have the structures `RevoluteJoint, PrismaticJoint, FixedJoint, SphericalJoint` that can be created and then converted to an `ImpulseJoint` or `MultibodyJoint`.


sebcrozet:
It is very interesting to see that this PR removes more lines than it adds: 12,643 removed lines vs. 9,869 added 😮 
This is mostly due to the fact that removing the position-based constraint solver, and merging all joints into a single 6-dofs joints, helped us get rid of a **lot** of special-cases (we hade one constraint type for each joint, for example).

PR #270, opened 2022-01-08
fix the parameter used for local center of mass

tjamaan:
In `rapier2d::pipeline::physics_pipeline::PhysicsPipeline::interpolate_kinematic_velocities()`:
The code that handles `RigidBodyType::KinematicVelocityBased`: `rb_vel.integrate()` expects a local center of mass which is already in `rb_mprops.local_mprops.local_com`.

### DIFF
--- a/src/pipeline/physics_pipeline.rs
+++ b/src/pipeline/physics_pipeline.rs
@@ -462,7 +462,7 @@ impl PhysicsPipeline {
                         &rb_pos.position,
                         // NOTE: we don't use the `world_com` here because it is not
                         //       really updated for kinematic bodies.
-                        &(rb_pos.position * rb_mprops.local_mprops.local_com),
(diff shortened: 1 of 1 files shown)


src/pipeline/physics_pipeline.rs
```diff
@@ -462,7 +462,7 @@ impl PhysicsPipeline {
                         &rb_pos.position,
                         // NOTE: we don't use the `world_com` here because it is not
                         //       really updated for kinematic bodies.
```
sebcrozet, line 464: This comment is no longer relevant after your change. Let’s remove it:

```suggestion
```

sebcrozet (changes requested):

sebcrozet (approved):

sebcrozet:
Thanks!

PR #274, opened 2022-01-11
add defaults for the several structs

### LINKED ISSUE #218: Implement Default for collection types
Currently RigidBodySet, ColliderSet, JointSet, Arena, Coarena and probably other collections provided by rapier do not implement the Default trait.

This is somewhat annoying, because it prevents the use of #[derive(Default)] and other Default based conveniences for structs that consist of rapier's collections.

It is also non-idiomatic, given that all collections in std implement Default.

It would be nice if Default was implemented for all collections which support creating an empty collection with no values.

besok:
Fix #218 
Just added several defaults methods

### DIFF
--- a/src/data/arena.rs
+++ b/src/data/arena.rs
@@ -52,9 +52,15 @@ pub struct Index {
     generation: u32,
 }
 
-impl IndexedData for Index {
+impl Default for Index {
     fn default() -> Self {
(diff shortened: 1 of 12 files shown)


sebcrozet:
Thanks! I rebased the branch onto master, and ran `cargo fmt`.

besok:
Thanks!

PR #286, opened 2022-01-31
Fix rigid bodies falling asleep when linear/angular threshold is negative

funmaker:
When a rigid body is created with `.can_sleep(false)`, the builder sets its activation thresholds to -1. However when the actual check is being made in island manager, the value of thresholds gets squared and sign is lost, allowing body to fall asleep.


sebcrozet:
Thanks!

PR #291, opened 2022-02-24
Start to Load World State

hayesgm:
This patch starts to load world state for debugging. The next step is to make sure that deserialization exactly matches the format of world.takeSnapshot().


sebcrozet:
Thanks!

PR #292, opened 2022-02-24
Derive Debug for IntegrationParameters

CleanCut:
Derive `Debug` for `IntegrationParameters` as per [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits).

I wanted to print out the value of the `IntegrationParameters` struct that `bevy_rapier2d` begins with, but it didn't implement `Debug`.


sebcrozet:
Thanks!

PR #313, opened 2022-04-20
Fix mass-properties update after collider change

sebcrozet:
This PR ensures that the mass-properties of a rigid-body are recomputed after modifying one of its attached colliders.
These mass-properties are recomputed from the colliders, as well as the newly stored additional mass-properties given to the rigid-body during its creation with the `RigidBodyBuilder`.


sebcrozet:
The CI failure will be addressed in the follow-up PR https://github.com/dimforge/rapier/pull/314.

PR #319, opened 2022-04-29
Allow collider modifications to wake-up their parent rigid-body

sebcrozet:
Until now, modifying a collider will not result in its parent being woken up, unless the user does so explicitly. Therefore, modifying, e.g., the collision groups of a collider may not have an immediate impact on the existing contacts because these contacts would not be recomputed if the rigid-body isn’t awaken.

In order to reduce the risk of the user forgetting to manually wake-up the collider’s parent, this PR add to colliders the ability to wake-up their parent automatically (at the next timestep).


sebcrozet:
Mmh, this actually isn’t the best way to handle this. A better approach is to make sure the narrow-phase properly recomputes contacts when the collider changed.

PR #327, opened 2022-05-14
fixed kinematic bodies not waking dynamic bodies on contact

gavlig:
In my project i added collision to camera so i would be able to push around dynamic physics and see how it behaves. The problem was that contact between kinematic and dynamic bodies didnt wake dynamics, so this is how i fixed that.


sebcrozet:
Thank you for this PR. This has been fixed on master during the sleeping system/solver rework.

PR #328, opened 2022-05-14
fixed dynamic bodies not waking when their joint motor changed

gavlig:
When wheels on my vehicle stop spinning they might fall asleep in rapier and using motors didnt wake them up so this is my fix

### DIFF
--- a/src/dynamics/joint/generic_joint.rs
+++ b/src/dynamics/joint/generic_joint.rs
@@ -206,6 +206,8 @@ pub struct GenericJoint {
     ///
     /// Note that the mostor must also be explicitly enabled by the `motors` bitmask.
     pub motors: [JointMotor; SPATIAL_DIM],
+    /// The flag indicating that motor settings were changed from outside
+    pub motor_changed: bool,
 }
 
 impl Default for GenericJoint {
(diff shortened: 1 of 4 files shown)


src/dynamics/joint/generic_joint.rs
```diff
@@ -206,6 +206,8 @@ pub struct GenericJoint {
     ///
     /// Note that the mostor must also be explicitly enabled by the `motors` bitmask.
     pub motors: [JointMotor; SPATIAL_DIM],
+    /// The flag indicating that motor settings were changed from outside
+    pub motor_changed: bool,
```
Vrixyz, line 210: While I appreciate the "low number of lines" this PR ends up, this would add state to an otherwise "configuration" or "data oriented" struct, I'm not sure this is the right place to add that.

ThierryBerger (changes requested):
Thanks for your report and suggestion for a fix ! Additionally to my comment (for which I do not have a solution to suggest right now), a reproduction / test / example would help with tackling that issue.

In the meantime it should be possible to wake up the body manually.

sebcrozet:
Thank you for this PR. This has been fixed on master during the solver rework.

PR #351, opened 2022-06-30
Update ActiveEvents Rustdoc to Remove Reference to EventHandler::handle_intersection_event()

aunyks:
The current `ActiveEvents` docs mention a `handle_intersection_event()` function of the `EventHandler` trait that no longer exists. This change removes that reference from the `ActiveEvents` docs.


sebcrozet:
Thanks!

PR #375, opened 2022-08-04
Reexport `BevyMaterial` alias

SludgePhD:
This type is used in the `TestbedPlugin` trait, and having to re-define the same type alias in every crate using it seems unnecessary.


sebcrozet:
Thanks!

PR #389, opened 2022-08-25
Update `bevy_egui`

SludgePhD:
(also fixes a build error introduced in https://github.com/dimforge/rapier/commit/68d250f0ad80bb2e5e10e43f2fd1b8824e2a1216 so that the examples build again)


sebcrozet:
Thanks!

PR #415, opened 2022-11-06
Add Rope Joints

Wolftousen:
Had made a game in Unity previously that had wrecking balls in it and for those I needed the Rope Joint to help keep the wrecking ball inline with the chain since a series of Revolute Joints has a "springy" behavior when a weight is attached to the end.

Reference [demo](https://piqnt.com/planck.js/RopeJoint) for Box2d Rope Joint.

Added both 2d and 3d implementations


marwie:
Same here!

sebcrozet:
Looking great, thanks @Wolftousen!

QuickMick:
i think, this and spring joints are missing in the documentation, isn't it?

sebcrozet:
@QuickMick This is in the [automatically generated doc](https://docs.rs/rapier3d/latest/rapier3d/dynamics/index.html) but not yet in the user-guide.

PR #424, opened 2022-12-04
KinematicCharacterController: early break in move_shape in case of no update.

PieroV:
Hi! I think the `while` in `move_shape` could be early-broken if no improvement is done in an iteration: if I understand correctly, if the character has not moved in an iteration, it will not in the next one, too.
However, all the collision queries will be run, consuming time.
But I am not sure this makes sense, since `kinematic_friction_translation` is also updated during iterations.

### DIFF
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -195,6 +195,7 @@ impl KinematicCharacterController {
         self.check_and_fix_penetrations();
 
         let mut translation_remaining = desired_translation;
+        let mut translation_before = result.translation;
 
         // Check if we are grounded at the initial position.
         let grounded_at_starting_pos = self.detect_grounded_status_and_apply_friction(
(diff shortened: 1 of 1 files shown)


PieroV:
@sebcrozet hi, you asked me to send a simple example of the problem that patch should solve on Discord, so here it is :slightly_smiling_face: : [simpletest.zip](https://github.com/dimforge/rapier/files/10202366/simpletest.zip).

Please notice also that I would expect grounded to become `true` eventually, but it does not (both with and without my patch). Maybe I can open an issue for that, unless I found a way to solve the problem.

sebcrozet:
Triaging. Thank you for this PR! I’m closing it since the provided zip does already stop after 2 iterations on the current `master` (and the grounded not being `true` appears to be fixed too). The proposed fix might have undesired side-effects if the solver do not move the object but still redirects the `translation_remaining` for the next iteration.

PR #429, opened 2022-12-21
Adds methods to retrieve forces added by the user.

fabriceci:
My first rust commit, a very small addition to get the strengths/torques added by the user.


fabriceci:
@sebcrozet The error comes from the file I did not modify, I get the same thing by running `cargo fmt --check".

There is a formatting problem in:

- examples2D/all_examples2.rs
- examples2D/rope_joints2.rs
- examples3d/rope_joints3.rs

sebcrozet:
Thanks!

PR #440, opened 2023-01-23
Allow customising the time until sleep for a rigid body

fabriceci:
Currently the time before a body falls asleep cannot be changed and is set to 2 seconds. This PR allows this value to be changed in the `RigidBodyActivation` of the body.


sebcrozet:
Thanks!

Issue #418, opened 2022-11-16
Offset of KinematicCharacterController is not sustained when desired translation points partially towards ground


kristiansvalland:
Hi, this is the first time I am posting an issue in this repo. I have encountered something I believe is a minor bug in the way the KinematicCharacterController updates the transformation under certain circumstances. I am using the bevy 3D plugin.

## Situation
In the code below demonstrating the problem, we have the following sequence of actions

1. The "character" spawns some height above the ground and due to gravity falls towards the ground
2. The character eventually reaches the ground, upon which is rests with the expected offset off from the ground.
3. After a delay of 2 seconds, a horizontal component is added to the speed of the character.
4. Together with gravity, which continuously tries to add a vertical component, the desired translation points partially down towards the ground.
5. This eventually leads to the character moving slightly towards the ground, now with it's "offset layer" overlapping with the ground.
6. Not shown in this example, this leads to the character intermittently becoming stuck.

The demonstration of the problem is printed via `warn` messages in the `read_result_system` function.

## Why it matters
As far as I understand from the documentation, and also based on experience from moving the character around, when the offset is too small, the character can inexplicably get stuck.

## Example code
``` rust
use bevy::prelude::*;
use bevy_rapier3d::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugin(RapierPhysicsPlugin::<NoUserData>::default())
        .add_startup_system(setup_physics)
        .add_system(gravity_movement_system)
        .add_system_to_stage(CoreStage::PostUpdate, read_result_system)
        .add_system_to_stage(CoreStage::PostUpdate, reconcile_character_velocity)
        .run();
}

fn setup_physics(mut commands: Commands) {
    /* Create the ground. */
    commands
        .spawn()
        .insert(Collider::cuboid(20.0, 20.0, 0.1))
        .insert_bundle(TransformBundle::from_transform(Transform::from_xyz(
            0.0, 0.0, -0.1,
        )));

    // Add KinematicCharacterController controlled body with a capsule shape
    commands
        .spawn()
        .insert(RigidBody::KinematicPositionBased)
        // Capsule of total height 2.0
        .insert(Collider::capsule_z(0.7, 0.3))
        .insert(KinematicCharacterController {
            offset: CharacterLength::Relative(0.01),
            up: Vec3::Z,
            ..default()
        })
        .insert(MySpeed::default())
[...]

PR #446, opened 2023-01-26
Make KinematicCharacterController move along obstacles, respect offset more and fix false positives in grounded detection

### LINKED ISSUE #426: KinematicCharacterController movement with surface normal corrections
One tedious issue that I encounter with the KinematicCharacterController (in both 2d and 3d) is that when I press my character against an object sideways, it no longer slides down, even though it is part of the `translation`. Ideally the velocity is clamped against the normal of surface it collides with.

Or perhaps I am using the api wrong?

janhohenheim:
- Fixes #426 
- Fixes https://github.com/dimforge/bevy_rapier/issues/301 modulo the documentation aspect
- Touched #418 and fixed it, but reverted because of jittering issues, see discussion.
- Fixes snapping to ground not triggering sometimes as described in https://github.com/dimforge/bevy_rapier/issues/301
- Fixes horizontal offset being mostly ignored and some instances of vertical offset being ignored
------
PR contents:
- Move stairs check before slope check to deal with some stopping along stairs
- Move along surface after stairs and slope checks
- Clean up some code, remove duplications
- Take offset into account in some places it was left out
- Fix ground being erroneously detected when jumping next to a wall composed of multiple triangles in 3D.

### DIFF
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -338,9 +338,7 @@ impl KinematicCharacterController {
                 ) {
                     // Apply the snap.
                     let snap_distance = hit.toi - offset;
-                    if snap_distance.abs() > 1.0e-5 {
-                        result.translation -= *self.up * snap_distance;
-                    }
+                    result.translation -= *self.up * snap_distance;
                     result.grounded = true;
                     return Some((hit_handle, hit));
                 }
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -232,7 +232,8 @@ impl KinematicCharacterController {
                 filter,
             ) {
                 // We hit something, compute the allowed self.
-                let allowed_dist = toi.toi - (toi.normal1.dot(&translation_dir)).abs() * offset;
+                let allowed_dist =
+                    (toi.toi - (-toi.normal1.dot(&translation_dir)) * offset).max(0.0);
                 let allowed_translation = *translation_dir * allowed_dist;
                 result.translation += allowed_translation;
                 translation_remaining -= allowed_translation;
@@ -263,7 +264,7 @@ impl KinematicCharacterController {
                         self.handle_slopes(&toi, &mut translation_remaining, offset)
                     {
                         translation_remaining = translation_on_slope;
-                    } else if allowed_dist < 1.0e5 {
+                    } else if allowed_dist.abs() < 1.0e5 {
                         // No slopes or stairs ahead, but we didn't move yet; try to move along obstacle.
                         let allowed_translation = subtract_hit(translation_remaining, &toi, offset);
                         result.translation += allowed_translation;
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -499,20 +499,16 @@ impl KinematicCharacterController {
         let angle_with_floor = self.up.angle(&hit.normal1);
         let climbing = self.up.dot(&slope_translation) >= 0.0;
 
-        climbing
-            .then(|| {
-                (angle_with_floor <= self.max_slope_climb_angle) // Are we allowed to climb?
-                    .then_some(horizontal_translation)
-            })
-            .unwrap_or_else(|| {
-                // Are we allowed to slide?
-                if angle_with_floor >= self.min_slope_slide_angle {
-                    slope_translation
-                } else {
-                    horizontal_translation
-                }
-                .into()
-            })
+        if climbing {
+            // Are we allowed to climb?
+            (angle_with_floor <= self.max_slope_climb_angle).then_some(horizontal_translation)
+        }
+        // Are we allowed to slide?
+        else if angle_with_floor >= self.min_slope_slide_angle {
+            Some(slope_translation)
+        } else {
+            Some(horizontal_translation)
+        }
     }
 
     fn split_into_components(&self, translation: &Vector<Real>) -> [Vector<Real>; 2] {
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -413,7 +413,7 @@ impl KinematicCharacterController {
                             let normal1 = character_pos * m.local_n1;
                             let normal2 = -normal1;
 
-                            if normal1.dot(&self.up) <= -1.0e-5 {
+                            if normal1.dot(&self.up).abs() <= -1.0e-5 {
                                 println!("Grounded 1");
                                 grounded = true;
                             }
@@ -463,7 +463,7 @@ impl KinematicCharacterController {
                         for m in &manifolds {
                             let normal = character_pos * m.local_n1;
 
-                            if normal.dot(&self.up) <= -1.0e-5 {
+                            if normal.dot(&self.up).abs() <= -1.0e-5 {
                                 for contact in &m.points {
                                     if contact.dist <= prediction {
                                         println!("Grounded 2");
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -328,20 +328,20 @@ impl KinematicCharacterController {
         result: &mut EffectiveCharacterMovement,
     ) -> Option<(ColliderHandle, TOI)> {
         if let Some(snap_distance) = self.snap_to_ground {
-            let snap_distance = snap_distance.eval(dims.y);
             if result.translation.dot(&self.up) < 1.0e-5 {
-                let offset = self.offset.eval(dims.y);
+                let snap_distance = snap_distance.eval(dims.y);
                 if let Some((hit_handle, hit)) = queries.cast_shape(
                     bodies,
                     colliders,
                     character_pos,
                     &-self.up,
                     character_shape,
-                    snap_distance - offset,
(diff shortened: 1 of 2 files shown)


janhohenheim:
Don't merge this quite yet, there's a tiny bug with the offset / snapping to ground.

janhohenheim:
Update: done done now :)

janhohenheim:
@sebcrozet PR should pass checks now

src/control/character_controller.rs
```diff
@@ -473,57 +474,41 @@ impl KinematicCharacterController {
...
-
-        // Check if there is a slope we can climb.
+        offset: Real,
+    ) -> Option<Vector<Real>> {
+        let [vertical_translation, horizontal_translation] =
+            self.split_into_components(translation_remaining);
+        let [vertical_slope_translation, horizontal_slope_translation] =
+            [vertical_translation, horizontal_translation]
+            .map(|remaining| subtract_hit(remaining, hit, offset));
+        let slope_translation = horizontal_slope_translation + vertical_slope_translation;
```
sebcrozet, line 480: I see, this is the key to fix the vertical wall problem. I’m not sure why I decided to apply that complex `if` statement to determine the `sliding_translation_remaining`.

src/control/character_controller.rs
```diff
@@ -251,25 +246,32 @@ impl KinematicCharacterController {
...
+                    // No stairs, try to move along slopes.
+                    if let Some(translation_on_slope) =
+                        self.handle_slopes(&toi, &mut translation_remaining, offset)
+                    {
+                        translation_remaining = translation_on_slope;
+                    } else {
+                        // No slopes or stairs ahead; try to move along obstacle.
+                        let allowed_translation = subtract_hit(translation_remaining, &toi, offset);
+                        result.translation += allowed_translation;
+                        translation_remaining -= allowed_translation;
```
sebcrozet, line 270: I may be missing something but this looks too optimistic. Not having any slope or step ahead does not imply that we can move along the obstacle by the complete `allowed_translation` since there could be other obstacles ahead.
In addition, doesn’t that imply that this `allowed_translation` would be applied twice? (It was already applied at line 238).

sebcrozet (changes requested):
Thank you for this PR!
While this fixes a few issues, it introduces new ones. In particular:
- Sometimes the autostepping on the 3D example doesn’t work.
- Sometimes the character does not stabilize on top of the jointed platform on the 2D example:

https://user-images.githubusercontent.com/1734958/216780948-4b3e55b1-0f5b-4dd5-8967-772e0868da49.mp4

I’m not sure what change exactly caused these issues.
See also the comments bellow.

I haven’t reviewed the changes to `handle_stairs` yet.

src/control/character_controller.rs
```diff
@@ -339,7 +339,10 @@ impl KinematicCharacterController {
                     filter,
                 ) {
                     // Apply the snap.
-                    result.translation -= *self.up * (hit.toi - offset).max(0.0);
+                    let snap_distance = hit.toi - offset;
+                    if snap_distance.abs() > 1.0e-5 {
```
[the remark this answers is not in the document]
janhohenheim: That was to combat jittering, but you're right, it's not needed in this case.
Reverted in [`6c5713a` (#446)](https://github.com/dimforge/rapier/pull/446/commits/6c5713aa0c8317e2ac428a546ae1ff0e1445e1e1)

src/control/character_controller.rs
```diff
@@ -295,19 +297,17 @@ impl KinematicCharacterController {
             }
         }
 
-        // If needed, and if we are not already grounded, snap to the ground.
-        if grounded_at_starting_pos {
```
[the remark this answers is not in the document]
janhohenheim: Good catch, reverted in [`b461d02` (#446)](https://github.com/dimforge/rapier/pull/446/commits/b461d028df09f8715646ca476e30693121777f31)

src/control/character_controller.rs
```diff
@@ -372,7 +379,7 @@ impl KinematicCharacterController {
             .compute_aabb(character_pos)
             .loosened(prediction);
 
-        let mut grounded = false;
```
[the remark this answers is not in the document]
janhohenheim: Huh, didn't think about that. You're right, reverted in [`7a2759c` (#446)](https://github.com/dimforge/rapier/pull/446/commits/7a2759c52feae6379da51fbf44584ea2c0b76c0c)

src/control/character_controller.rs
janhohenheim: Yep, together with the block prefaced by `// No slopes or stairs ahead; try to move along obstacle.`. I was quite happy to be able to simplify the `if` construct 😄

janhohenheim:
Thanks for the review, I'll fix the issues and ping you once I'm done :)

src/control/character_controller.rs
janhohenheim: Hmm, I see. For the fix, I noticed that often it's not a second time, but a first time. The line you mention stops the movement completely in some cases, e.g. when trying to jump up while moving up the wall on the right. How about we run this code only when it would be the first we move along the obstacle? See [`760ea8d` (#446)](https://github.com/dimforge/rapier/pull/446/commits/760ea8d498c7dfc99d59342c8ec492de193ffed9)

janhohenheim:
[`c9462e0` (#446)](https://github.com/dimforge/rapier/pull/446/commits/c9462e0818aca202f63fa3c931e3a119aa252a55) fixes the jittering, but causes the character to violate the offset in some cases. This is not a regression since the main branch suffers same issue.
image

janhohenheim:
@sebcrozet I addressed your concerns, ready for another review :)

morbidbark:
I just checked out your branch to see if it would fix some issues I was having with the KinematicCharacterController(offset not being respected and getting stuck to walls). It did fix those problems, but introduced a few others. Lot's of jitter and very  odd interactions with moving platforms.

janhohenheim:
@morbidbark that's a bummer. I did see some jitter as well, but it's hard to tell how much of that was already present and not noticeable because before the character didn't move at all in many of these situations. I tried looking further into it, but I'm at a loss here.

morbidbark:
So I played around with it a little more today and found that when standing still on solid ground my CharacterController is constantly switching between the grounded and falling state. However, when I ran the character_controller2 example and monitored the grounded attribute on EffectiveCharacterMovement it appeared to be stable. It looks like my problems are probably just user error.

morbidbark:
I was able to verify that my jittering problems were due to user error. I had frames where my "gravity" was not acting on my controller causing it to appear ungrounded. The issues I was having with the moving platforms is still there, but this also happens on the latest release so not a problem with this PR. Sorry for any confusion.

janhohenheim:
Glad to hear 😄

sebcrozet:
Hey @janhohenheim!
Sorry for taking so long for getting back to this PR, I’ve been on vacation.
Thank you for your latest changes.


Taking a closer look at `handle_stairs` reminded me of why it was run after `handle_slopes`. If stairs are handled first, climbing slopes (the ones with an angle smaller than the `max_slope_climb_angle`) will actually be handled partially (or completely if moving at low speed) by the stair handling code. Which I guess is generally fine, just something to keep in mind.

Now, the new slope handling code does the following:
```rust
        if climbing {
            // Are we allowed to climb?
            (angle_with_floor <= self.max_slope_climb_angle).then_some(horizontal_translation)
        }
```
the comparison is incorrect here as it reads "if the floor is angles by less than `max_slope_climb_angle`, prevent climbing the slope", which is the opposite of what `max_slope_climb_angle` means. This has two consequences:
- When auto-step is enabled, `climbing` will always be `false` when `angle_with_floor <= self.max_slope_climb_angle` because that climbing has already been dealt with by the stair handling. The reason why the user cannot climb steep slopes in this PR is only because of  gravity: if there isn’t enough movement uphill to overcome gravity, the player won’t be able to go up.
- When auto-step is disabled, then the player will no longer be able to climb small slopes because that test is reversed.

What if we fixed that check with a `>=` instead of a `<=`? Well, in that case, sliding against walls no longer works because it’s prevented by this threshold.

I think we can have the `>=` case return `vertical_translation`, and `None` otherwise. This will prevent strong horizontal motion push through the slope, letting the user climb due to the indirect vertical movement that emanate from this horizontal motion minus the diagonal contact normal.

In addition, having `handle_slopes` return `Some(slope_translation)` is equivalent to returning `None` because of the `None` case results in a `subtract_hit` equivalent to the ones applied to get `slope_translation`. This lets us simplify the code slightly.

I don’t want to delay this PR further so I’ll make a commit for these observations and merge the PR. We can still improve it further later if these suggestions still have issues.

sebcrozet (approved):

sebcrozet:
Fantastic work @janhohenheim, thanks again!

janhohenheim:
@sebcrozet no worries, hope you enjoyed your vacation :)
Thanks for bringing these points up, I'll see what I can do about them next week

PR #461, opened 2023-03-11
Prevent parallel wasm builds

bugQ:
Attempting to target wasm with the parallel feature on fails because `rayon` does not natively support it.

Parallelism on WebAssembly is technically achievable through the Web Workers API, but until such functionality is stable, this should provide a better error message.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,6 +31,9 @@ extern crate num_traits as num;
 // #[macro_use]
 // extern crate array_macro;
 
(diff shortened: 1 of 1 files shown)


Vrixyz:
I believe since https://github.com/rayon-rs/rayon/pull/1110 ; rayon has a synchronous fallback to work on web ; needs verification though.

bugQ:
Awesome, sounds like this PR is not needed then.

PR #468, opened 2023-03-21
Add Ability to Control Side-to-Side Grip / Forward Grip of Raycast Vehicle Wheels

aunyks:
These changes add a struct item to `Wheel` and `WheelDesc` to allow users to control how much lateral grip a raycast vehicle's wheel can have. Lower values [0-1) give an effect similar to sliding or drifting. Before these changes, [the effective side friction stiffness was `1.0`](https://github.com/dimforge/rapier/blob/master/src/control/ray_cast_vehicle_controller.rs#L512).

They also make `Wheel`'s `friction_slip` property public so it can be changed after wheel creation.

Also fixed a small spelling error in some doc comments in the raycast vehicle source file (`his` -> `has`).

### DIFF
--- a/examples3d/vehicle_controller3.rs
+++ b/examples3d/vehicle_controller3.rs
@@ -44,7 +44,7 @@ pub fn init_world(testbed: &mut Testbed) {
     ];
 
     for pos in wheel_positions {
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thanks!

PR #470, opened 2023-03-21
#465 Add is_sliding_down_slope field to EffectiveCharacterMovement

### LINKED ISSUE #465: KinematicCharacterController min_slope_slide_angle does not send any information to output when slope angle is larger than min_slope_slide_angle
Why is this an issue?
The KinematicCharacterController has a field min_slope_slide_angle that when set, it should make entities that are on steeper slopes to slide down. While the handle_slopes function here https://github.com/dimforge/rapier/blob/c85a833d3a2e4c66024627a9745f033fbbee283a/src/control/character_controller.rs#L475
does do that, it does not send any information to the consumer that it is sliding down a slope. As a consumer, there needs to be a way to know if an entity is currently sliding down a slope in order to maintain things like gravity acceleration.

Potential fix:
Add an extra boolean flag such as `is_sliding_down_slope` to the KinematicCharacterControllerOutput when https://github.com/dimforge/rapier/blob/c85a833d3a2e4c66024627a9745f033fbbee283a/src/control/character_controller.rs#L498 is reached.

Jeff425:
Resolves #465 
This change adds the `is_sliding_down_slope` field to the EffectiveCharacterMovement output struct for `move_shape` KinematicCharacterController function

This field will be true when the entity is sliding down at least one slope.

From my testing, this field will be true when sliding into a wall. So if the consumer does not want a player to move while sliding, make sure to check for that edge case.

### DIFF
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -135,6 +135,8 @@ pub struct EffectiveCharacterMovement {
     pub translation: Vector<Real>,
     /// Is the character touching the ground after applying `EffectiveKineamticMovement::translation`?
     pub grounded: bool,
+    /// Is the character sliding down a slope due to slope angle being larger than `min_slope_slide_angle`?
+    pub is_sliding_down_slope: bool,
 }
 
 impl KinematicCharacterController {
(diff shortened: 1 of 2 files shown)


Jeff425:
@sebcrozet Mind assigning a reviewer? Seems like there's a few people interested in this change

sebcrozet (approved):

sebcrozet:
Thanks!

PR #474, opened 2023-03-26
Add a predicate to the DebugRenderBackend to filter out objects that are being rendered

### LINKED ISSUE #398: Allow to filter groups in `DebugRenderPipeline`
Currently, it's only possible to filter what kind of item we want to display, but not filter by group. Here is my use case, I have two groups: `OUTSIDE_WORLD` and `INSIDE`. So when I'm entering the `INSIDE` state, I don't want the items from the `OUTSIDE_WORLD` group to be displayed (but I don't remove them as it would take too much time to re-pop them, I only do it for `INSIDE` elements).

Do you think such a feature could be added?

Also: thanks a lot for your crates, they're really incredible to use!

PS: If you think it's a nice feature to add, I can take a look at adding it myself if you want?

sebcrozet:
Fix #398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Modified
 - Make `Wheel::friction_slip` public to customize the front friction applied to the vehicle controller’s wheels.
+- Add the `DebugRenderBackend::filter_object` predicate that can be implemented to apply custom filtering rules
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Nice, thanks! Could you add an example for it as well please?

PR #476, opened 2023-03-30
Fix doc-links for `ColliderBuilder` and `RigidBodyBuilder`

MrGunflame:
Fix the doc-links to `ColliderBuilder` and `RigidBodyBuilder` from their respective types.


sebcrozet:
Thanks!

PR #478, opened 2023-03-30
Fix out of bounds access

### LINKED ISSUE #477: Panic when changing rigid body types or when disabling rigid bodies that are colliding with each other in bevy plugin w/ fixed time step.
The error message is that an index out of bounds error is happening here in `interaction_groups.rs` line 440:
```rs
let mask1 = if !is_fixed1 { self.body_masks[i1] } else { 0 };
let mask2 = if !is_fixed2 { self.body_masks[i2] } else { 0 };
```

A simple fix could be verifying i1 and i2 are within self.body_masks.len() - I'll test that to see.

AnthonyTornetta:
Resolves #477 by checking index is within bounds before accessing data at that index


sebcrozet:
Thank you for this PR. This fix isn’t ideal as it result in some contacts being ignored by the constraints solver.
The case that is being checked here shouldn’t happen, so something must be wrong with disabling bodies as you mentioned in #477.

PR #481, opened 2023-04-02
should allow grounding with no movement counter to the up direction

ChrisJanssens:
It looks like we are taking the dot product of the movement and up direction here to ensure the vector is negative to the up direction. Is this necessary? I was using grounded to determine when my character controller can "jump". Allowing this value to be zero or lower allows my character controller lateral movement along a collider that is on the opposed side of my character to that of the up direction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@
   violation recovers over a few steps instead of catapulting the bodies.
 - Python: `DynamicRayCastVehicleController.update_vehicle` now excludes the chassis body from
   the suspension raycasts by default (their origins sit on its own collider).
+- The character controller's snap-to-ground now triggers on any movement that isn't upwards,
+  including movements exactly orthogonal to the up vector, instead of requiring a downward
+  motion larger than an arbitrary epsilon. Purely lateral movement along the ground no longer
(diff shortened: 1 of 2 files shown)


ThierryBerger (approved):

Vrixyz:
in practice you probably still want to apply a gravity toward the "ground" ? So I'm not sure this matters much ; but an epsilon isn't needed I think indeed.

Vrixyz:
While I think this makes sense, more context about this code can be found here:
- https://github.com/dimforge/rapier/pull/446. 
  - and especially https://github.com/dimforge/rapier/pull/446/commits/7e1fb6f8770ee41a2f5a66ee2c8b15e61ddac6a4

The commit message is explicitly about forbidding exactly horizontal movement, I'm not sure wheteher it's fully thought out, or was done only to follow the epsilon "mindset" which was already here before. Tagging @janhohenheim if you're around and able clarify ? 🥺

janhohenheim:
@Vrixyz I remember doing that, but I don't remember my reasoning. 
If it's inconvenient, feel free to revert it :)

Vrixyz:
I still agree with this change, but this only adds a fix when snapping is enabled, so we'll need to gather more information about it before validating.

This PR https://github.com/dimforge/rapier/pull/715 explores alternatives (or additions)

sebcrozet:
Thank you!

PR #484, opened 2023-04-24
Prevent testbed.rs from crashing when pressing D

stefnotch:
Steps to reproduce the bug before this fix:
1. Start testbed
2. Switch to character controller example
3. Press D
4. Crash


sebcrozet:
Thanks!

PR #495, opened 2023-06-29
Remove usage of magic numbers in DynamicRaycastVehicleController

JohnTheCoolingFan:
Usage of magic numbers is discouraged in rust in favor of enums. I've replaced magic numbers that denote axes with an enum that can be converted to an integer when needed. The enum uses less bytes and does not allow invalid states.


sebcrozet:
Triaging. Thank you for this PR. We already have a bunch of enums for axis so I wouldn’t want to add yet another one specific to the linear axis only for the vehicle controller. Closing this for now. We can revisit this if requested again by the community.

PR #505, opened 2023-07-15
`user_data` for joints

mrDIMAS:
Removes inconsistency with other entities, such as rigid bodies and colliders - they all have `user_data` field. 

## Motivation

This field is usually used to store a pointer/handle to game/engine/etc entities associated with a rigid body/collider/joint. And the lack of `user_data` field causing troubles in Fyrox - I forced to use a hash map like `HashMap<Handle<Stuff>, JointHandle>` to bypass this issue, which has both performance and usability impact.


sebcrozet:
Thanks!

Issue #506, opened 2023-07-26
 Add serde derives for KinematicCharacterController


PaulL48:
KinematicCharacterController appears to be a good candidate to be Serialize/Deserialize. This would require adding serde derives for the following structs.

- CharacterLength
- CharacterAutostep
- KinematicCharacterController

PaulL48:
I've got the changes tested and ready for a PR. I rarely contribute to OS projects so I just wanted to follow the guideline and create an issue to be assured that everything is OK before creating the PR.

sebcrozet:
The corresponding PR https://github.com/dimforge/rapier/pull/508 was merged a while ago. We can close this.

PR #508, opened 2023-08-02
Add serde derives to structures in the control module

PaulL48:
Adding feature-gated serialize/deserialize derives for the following structs:

- CharacterLength
- CharacterAutostep
- KinematicCharacterController

Resolves issue #506


sebcrozet:
Thanks!

PR #511, opened 2023-08-10
fix lock_translations & lock_rotations doesn't work both ways

Triuman:
rigidbody.lock_translations and lock_rotations methods were able to lock but not unlock.


sebcrozet:
Thanks!

PR #515, opened 2023-08-24
feat/Color of disabled object will change

### LINKED ISSUE #489: Debug renderer: indicate disabled rigidbodies and colliders
The debug renderer changes the color when an object is sleeping, but not when it is disabled.

Showing a disabled collider like normal can be confusing when nothing interacts with it. I was stuck debugging collider groups instead of why it was incorrectly disabled.

Soham1803:
Fixes #489


sebcrozet:
Thanks!

PR #518, opened 2023-08-28
deps: Remove unused `indexmap` dep.

waywardmonkeys:
This was only enabled in the `enhanced-determinism` feature, but it wasn't actually used in these crates.


sebcrozet:
Thanks!

PR #527, opened 2023-09-27
Calculate world center of mass from local com in debug renderer

Aceeri:
Kinematic bodies that were moved after creation don't have their world_com mass properties re-calculated. This isn't a complete solution since I think ideally the `rb.mrprops.world_com` should be updated in that case, but fixes the debug renderer for now.


sebcrozet:
Thanks!

PR #529, opened 2023-09-30
Use the absolute value of the up attribute when calculating up_extent and side_extent

Linentio:
Fixes https://github.com/dimforge/bevy_rapier/issues/413

I think that we should use the absolute value of the `up` attribute of a `KinematicCharacterController` when calculating the value of the `up_extent` and `side_extent` variables.
Concerning the `up_extent` variable, if you make the `up` attribute point downward, parry will make the app crash because of the `assert` in the `loosened` function of the `AABB` struct since the amount of loosening would be negative.
And concerning the `side_extent` variable, the value would not be the same if the `up` attribute points upward or downward.


sebcrozet:
Thanks!

PR #531, opened 2023-10-04
add bug

Ughuuu:
Made a PR so you can view the bug I am experiencing. Here is the image of how the example looks like:

image

What is happening here, I am calling the `parry::query::contact` and getting No collision, even though the shapes are obviously colliding.

From my investigation, the code goes to [epa2.rs](https://github.com/dimforge/parry/blob/af38912c942afd260e3d65931c750e57beb7c4b0/src/query/epa/epa2.rs#L161)

And then it stops at:

```
if niter > 10000 {
    return None;
}
```

It seems to be adding the same face over and over 1000 times to the heap and removing it, and then it stops.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -67,7 +67,7 @@
             "name": "Debug executable 'all_examples2'",
             "cargo": {
                 "args": [
-                    "build",
+                    "run",
                     "--bin=all_examples2",
                     "--package=rapier-examples-2d"
                 ],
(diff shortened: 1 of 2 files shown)


Ughuuu:
In the end bug only happens with capsule. So I ended up making the capsule from 2 circles and 1 square.

Vrixyz:
I couldn't compile this specific branch so I updated with master: https://github.com/Vrixyz/rapier/tree/add-bug_updated ; I can reproduce the bug.

PR #538, opened 2023-10-18
Fix enabling disabled collider

rafalh:
If collider was enabled and no other attributes were changed its collisions were not detected.
This was caused by `needs_broad_phase_update` function not handling the case of changed enabled status.
Fixes dimforge/bevy_rapier#435


bgalek (approved):

sebcrozet:
Thanks!

PR #560, opened 2023-12-02
Fix check for motor constraints on angular DoF

kvark:
It looks like `motor_axes` is just a mask of axes, similar to `locked_axes`. It shouldn't need to be shifted by `DIM` for these checks.


sebcrozet (approved):
Good catch. Looks like a bug introduced when `i` was iterating starting at 0 instead of `DIM` directly.

PR #569, opened 2023-12-17
add reapply_lost_speed to KinematicCharacterController

ntibi:
Adds `reapply_lost_speed: bool` to KinematicCharacterController

When set to true, the velocity lost by sliding will be reapplied to the corrected direction
so the character moves at the same speed wether it's sliding or not

### DIFF
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -295,7 +295,7 @@ impl KinematicCharacterController {
 
                 let hit_info = self.compute_hit_info(hit);
                 let pre_slide_translation_remaining = translation_remaining;
-                // Try to go up stairs.
+                // Try to go upstairs.
                 if !self.handle_stairs(
                     bodies,
                     colliders,
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -128,6 +128,9 @@ pub struct KinematicCharacterController {
     pub offset: CharacterLength,
     /// Should the character try to slide against the floor if it hits it?
     pub slide: bool,
+    /// Should the speed lost because of collisions be applied to the final direction ?
+    /// Useless if sliding is not enabled
+    pub reapply_lost_speed: bool,
     /// Should the character automatically step over small obstacles? (disabled by default)
     ///
     /// Note that autostepping is currently a very computationally expensive feature, so it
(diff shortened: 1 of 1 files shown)


src/control/character_controller.rs
```diff
@@ -267,6 +272,17 @@ impl KinematicCharacterController {
...
+                    let diff =
+                        pre_slide_translation_remaining.norm() - translation_remaining.norm();
+                    if let Some((normalized_dir, _)) =
+                        // threshold has to be > subtract_hit hardcoded correction, if not, we get made up movement in the direction of the hit normal
+                        UnitVector::try_new_and_get(translation_remaining, 1.0e-4)
+                    {
+                        // reapply the lost speed (but in the corrected direction)
+                        translation_remaining += *normalized_dir * diff;
+                    }
+                }
```
Vrixyz, line 285: That condition should probably be in case there is no stairs ? otherwise we might end up intersecting with the stair. (and it might result in not going up the stair, or worse an infinite loop ?)

ntibi: indeed, but even with the reapply in the `if !self.handle_stairs`, reapplying the speed can be slightly flaky, especially in 3d

im turning the PR back into a draft and looking into it

MeFisto94:
I guess kind of reworking the controller to not have the speed loss in the first place is not really possible? Would that result in loads of loop iterations of the slope sweeps?

I am asking because technically this behavior of slowing down/speeding up on slopes is not what one would expect from a character controller, especially contradicting the "kinematic" keyword. Same as the concept of "friction", it should usually always be infinite or zero.

For slopes specifically, the expectation (subjectively) is that one can climb them at full speed and descend them at full speed, unless the angle exceeds the maximum angle, _then_ slipping is involved in some way.

My problem currently also is that I can't control which slopes to climb as the effect of the slope angles is too high, while also causing considerate CPU load (the loop that does the offset by the normal nudge and then raytraces again).

So, TLDR: This sounds like it solves problems of an actual dynamic character controller, not a kinematic one and properly getting that kinematic behavior may involve going back to the drawing board

sebcrozet:
Triaging. Thank you for this PR. I’m closing it since it’s been on draft for a while. If that remains a problem, please open an issue with a minimal scenes that would benefit from this.

PR #599, opened 2024-03-11
fix(user_changes): Fix RigidBodyType changed to Dynamic not added to active dynamic set

MaxCWhitehead:
I ran into what I believe is a bug in which if a rigid body is changed to dynamic from kinematic, it may never end up in active dynamic set. This caused some issues including gravity not being integrated.

I believe in user changes we should add body to dynamic active if its type changed to dynamic. (And is enabled / not sleeping, the existing conditions).

I added a test to reproduce the issue. The failure seems to only occur if pipeline is stepped (at least) once after creating kinematic body, before switching to dynamic. Without the change to user_changes.rs, both conditions will fail in test.

Please let me know if I am somehow using API wrong and this is not a bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fix
 
 - Fix crash when simulating a spring joint between two dynamic bodies.
+- Fix kinematic bodies not being affected by gravity after being switched back to dynamic.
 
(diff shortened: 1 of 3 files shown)


sebcrozet:
That makes sense. Thank you for the fix!

PR #604, opened 2024-03-23
feat: add RigidBody::predict_position_using_velocity

### LINKED ISSUE #601: Need predict_position_using_velocity() without forces
There is [predict_position_using_velocity_and_forces](https://docs.rs/rapier3d/latest/rapier3d/dynamics/struct.RigidBody.html#method.predict_position_using_velocity_and_forces). The problem is - if you have a vehicle driving on the road, and you want to estimate the next vehicle position, you don't want to assume that it's going to be pulled by gravity. I.e. it would be useful to have a prediction mode that assumes that the forces are in the equilibrium.

sebcrozet:
Fix #601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 - Fix crash when simulating a spring joint between two dynamic bodies.
 - Fix kinematic bodies not being affected by gravity after being switched back to dynamic.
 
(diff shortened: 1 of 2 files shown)


kvark:
Awesome, thank you for writing this!

PR #623, opened 2024-04-21
Fix one body joints limits not being flipped

### LINKED ISSUE #549: Dynamic -> Ground impulse joint limits are (not) flipped
Joint constraint solver has separate functions for dealing with Dynamic <-> Fixed/Kinematic bodies. In them, it assumes that the body1 of the joint is the static one, and if that's not the case, it flips the bodies around https://github.com/dimforge/rapier/blob/e9ea2ca10b3058a6ac2d7f4b79d351ef18ad3c06/src/dynamics/solver/joint_constraint/joint_constraint.rs#L271-L276
However it does not flip the limits so if two bodies have asymmetric limits(eg, 45deg to 0), they can jump around when one of them becomes Fixed.
I can provide reproduction but I think the problem should be evident from the lines of code mentioned above. I will include a video.

https://github.com/dimforge/rapier/assets/3338680/aa904259-2c03-4544-a1da-8e3bb6d29e29

I am not sure if simply flipping sign of limits would be enough to fix that(are angular limits symmetric like that?) and whenever this also affects multibody joints.

funmaker:
This PR attempts to fix #549

I am not 100% convenient in the correctness of this change, but it does fix the issue I was experiencing in my program. Especially the changes to `JointGenericVelocityOneBodyExternalConstraintBuilder`, since as far as I can tell this is used with multibodies, which I do not use. But I followed the same logic I did with `JointOneBodyConstraintBuilder`.

I also removed some outdated TODO comments that are no longer relevant I think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Fix kinematic bodies not being affected by gravity after being switched back to dynamic.
 - Fix regression on contact force reporting from contact force events.
 - Fix kinematic character controller getting stuck against vertical walls.
+- Fix joint limits/motors occasionally not being applied properly when one of the attached
+  rigid-bodies is fixed.
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


sebcrozet:
This is perfect, thank you @funmaker!

PR #638, opened 2024-05-27
Prefer individual `unexpected_cfgs` allow over global one

Urgau:
This PR replace the global `#![allow(unexpected_cfgs)]` over individual ones in `crates/*`, so that the main library still gets the benefits of the lint.

This PR also remove obsolete asmjs cfg, since that target was removed some time ago from Rust.

### DIFF
--- a/benchmarks2d/all_benchmarks2.rs
+++ b/benchmarks2d/all_benchmarks2.rs
@@ -31,7 +31,7 @@ fn demo_name_from_command_line() -> Option<String> {
     None
 }
 
-#[cfg(any(target_arch = "wasm32", target_arch = "asmjs"))]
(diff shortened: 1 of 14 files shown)


Vrixyz (approved):
Looks fine to me ; it's a pity we're losing the comment for context on the lint though

For the asmjs target drop, more context here: https://github.com/rust-lang/rust/pull/117338

Vrixyz (commented):
After playing a bit with it, I'm not sure if this makes sense to keep `unexpected_cfgs = "allow"` at all ?

sebcrozet:
@Urgau Thank you for this PR! What is the benefit of allowing the lint in individual `Cargo.toml` (8 times) instead of `lib.rs` (only 2 times)?

PR #646, opened 2024-06-05
Character controller: change solve_character_collision_impulses to take multiple CharacterCollision

Vrixyz:
https://github.com/dimforge/rapier.rs/pull/104 will have to adapt to this modification

### DIFF
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -798,7 +798,7 @@ impl KinematicCharacterController {
         filter: QueryFilter,
     ) {
         for collision in collisions {
-            self.solve_character_collision_impulse(
+            self.solve_single_character_collision_impulse(
                 dt,
                 bodies,
                 colliders,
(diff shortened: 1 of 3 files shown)


src/control/character_controller.rs
```diff
@@ -782,11 +782,40 @@ impl KinematicCharacterController {
...
+                filter,
+            );
+        }
+    }
+
+    /// For the given collision between a character and its environment, this method will apply
+    /// impulses to the rigid-bodies surrounding the character shape at the time of the collision.
+    /// Note that the impulse calculation is only approximate as it is not based on a global
+    /// constraints resolution scheme.
+    fn solve_character_collision_impulse(
```
sebcrozet, line 818: Just to make it extra obvious it solves a single collision:
```suggestion
    fn solve_single_character_collision_impulse(
```

sebcrozet (approved):
Thanks!

src/control/character_controller.rs
Vrixyz: Notice I changed its visibility to private, is that ok ? @sebcrozet

sebcrozet: Yes, making it private is good.

PR #647, opened 2024-06-05
Rework QueryPipeline update API to take less parameters

Vrixyz:
When working on https://github.com/dimforge/rapier.rs/pull/104 ; I noticed an outdated note on the userguide, which was also teasing future features.

<details><summary>Note details</summary>
<p>

at https://rapier.rs/docs/user_guides/rust/scene_queries

> Right now, the query pipeline takes the `IslandManager` and `RigidBodySet` as arguments. These two arguments may look superfluous, especially if the colliders are not attached to rigid-bodies. Another update method that only takes a `ColliderSet` as its only argument will be added in the future.

</p>
</details> 

This PR allows to remove that note, and makes up for a more powerful API.

On its downside, passing a generic is less explicit, I tried my best to guide the user with the documentation.

### DIFF
--- a/src/dynamics/ccd/ccd_solver.rs
+++ b/src/dynamics/ccd/ccd_solver.rs
@@ -4,7 +4,7 @@ use crate::geometry::{ColliderParent, ColliderSet, CollisionEvent, NarrowPhase};
 use crate::math::Real;
 use crate::parry::utils::SortedPair;
 use crate::pipeline::{EventHandler, QueryPipeline};
-use crate::prelude::{updaters, ActiveEvents, CollisionEventFlags};
+use crate::prelude::{query_pipeline_generators, ActiveEvents, CollisionEventFlags};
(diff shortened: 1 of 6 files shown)


src/dynamics/ccd/ccd_solver.rs
```diff
@@ -117,11 +117,8 @@ impl CCDSolver {
         narrow_phase: &NarrowPhase,
     ) -> Option<Real> {
         // Update the query pipeline.
-        self.query_pipeline.update_with_mode(
-            bodies,
-            colliders,
-            QueryPipelineMode::SweepTestWithPredictedPosition { dt },
-        );
+        self.query_pipeline
+            .update_with_mode(updaters::SweepTestWithNextPosition { bodies, colliders });
```
Vrixyz, line 121: should be `SweepTestWithPredictedPosition`

sebcrozet (approved):

PR #663, opened 2024-06-22
Fix `attempt to subtract with overflow` panic in `SAPRegion::update_after_subregion_removal()`

### LINKED ISSUE #593: Crash with `attempt to subtract with overflow` in `SAPRegion::update_after_subregion_removal()`
I am using rapier 0.18.0 through bevy_rapier2d 0.24.0 and noticed occasional `attempt to subtract with overflow` in `update_after_subregion_removal()` on subtract from `self.subproper_proxy_count`. I cloned rapier repo, added some debug and turns out that `axis.delete_deleted_proxies_and_endpoints_after_subregion_removal()` call returns 1 but `self.subproper_proxy_count` equals 0 so subtraction causes panic in debug build.

As quick workaround I changed the subtraction to `self.subproper_proxy_count = self.subproper_proxy_count.saturating_sub(num_deleted_proxies)` but it's out of my rapier knowledge to fix this properly.

I'm attaching log containing debug outputs and stack backtrace + diff with changes I made. I know how to reproduce this problem (~50% chance) with current state of my project so it's possible to provide additional logs if needed.
[log.txt](https://github.com/dimforge/rapier/files/14344796/log.txt)
[diff.txt](https://github.com/dimforge/rapier/files/14344813/diff.txt)

DiSaber:
This works around #593, although I'm unsure of the root cause of the issue.

### DIFF
--- a/src/geometry/broad_phase_multi_sap/sap_region.rs
+++ b/src/geometry/broad_phase_multi_sap/sap_region.rs
@@ -192,13 +192,23 @@ impl SAPRegion {
     pub fn update_after_subregion_removal(&mut self, proxies: &SAPProxies, layer_depth: i8) {
         if self.needs_update_after_subregion_removal {
             for axis in &mut self.axes {
-                self.subproper_proxy_count = self.subproper_proxy_count.saturating_sub(
-                    axis.delete_deleted_proxies_and_endpoints_after_subregion_removal(
+                let removed_count = axis
+                    .delete_deleted_proxies_and_endpoints_after_subregion_removal(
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thank you for this PR! I changed it a little bit to generate a (debug) log if that happens.
This needs to be investigated further, but at least I think it’s fine to disable the panic in the mean time.

I modified your message (replaced "fixed" by "works around") so that github doesn’t auto-close the linked issue (I want to keep the issue open to keep the bug in mind for future investigations).

PR #671, opened 2024-07-05
ci: cargo doc step

Vrixyz:


Vrixyz:
@sebcrozet Can you check the following error in ci and confirm if we can get rid of this mention ?

```rs
857 |     /// or [`Self::finalize`].
    |              ^^^^^^^^^^^^^^ the struct `Multibody` has no field or associated item named `finalize`
```

src/geometry/collider.rs
```diff
@@ -620,10 +620,22 @@ impl ColliderBuilder {
...
     /// Initialize a new collider builder with a capsule defined from its endpoints.
     ///
-    /// See also [`ColliderBuilder::capsule_x`], [`ColliderBuilder::capsule_y`], and
-    /// [`ColliderBuilder::capsule_z`], for a simpler way to build capsules with common
+    /// See also [`ColliderBuilder::capsule_x`], [`ColliderBuilder::capsule_y`],
+    /// and [`ColliderBuilder::capsule_z`],
+    /// for a simpler way to build capsules with common
     /// orientations.
     pub fn capsule_from_endpoints(a: Point<Real>, b: Point<Real>, radius: Real) -> Self {
         Self::new(SharedShape::capsule(a, b, radius))
```
sebcrozet, line 641: The code duplication just for the documentation is unfortunate. Perhaps we could just keep a single method, and have the documentation be something like:
```
See also [`ColliderBuilder::capsule_x`], [`ColliderBuilder::capsule_y`] (and `::capsule_z` in 3D only), for a ...
```

(no doc-link for `capsule_z`)

sebcrozet (approved):
Thank you for this PR! I confirm that the mention to`Self::finalize` can be removed.

PR #673, opened 2024-07-08
Fix QueryFilterFlags values having a bitshift too much

Vrixyz:
- Fixes https://github.com/dimforge/rapier.js/issues/275


sebcrozet (approved):
Thank you! Let’s add a `### Modified` entry to the changelog saying `Divided by two the value of each QueryFilterFlags variant so that the smallest one is 1 instead of 2. (Fixes a bug in rapier.js).

PR #675, opened 2024-07-09
Fix string serialization for broadphase multisap

Vrixyz:
- Context: https://github.com/dimforge/bevy_rapier/pull/560.

Without that, serialization to string fails. Interesting to note `bincode` works without that.

error log:

```txt
thread 'main' panicked at bevy_rapier3d/examples/serialization.rs:30:51:
Unable to serialize `RapierContext`: Error("key must be a string", line: 0, column: 0)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `serialization::print_physics`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

After more testing, I realize this crash doesn't happen every time 😮 ; around once in 10 runs 🤔.

The solution for this issue is more involved than I thought, serde json doesn't support non-string keys for `Hashmap`: https://github.com/serde-rs/json/issues/887 ; I'm not sure what's the path forward.

### DIFF
--- a/src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
+++ b/src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
@@ -93,8 +93,8 @@ pub struct BroadPhaseMultiSap {
     #[cfg_attr(
         feature = "serde-serialize",
         serde(
-            serialize_with = "parry::utils::hashmap::serialize_hashmap_capacity",
-            deserialize_with = "parry::utils::hashmap::deserialize_hashmap_capacity"
+            serialize_with = "crate::utils::serde::serialize_to_vec_tuple",
+            deserialize_with = "crate::utils::serde::deserialize_from_vec_tuple"
         )
     )]
     colliders_proxy_ids: HashMap<ColliderHandle, BroadPhaseProxyIndex>,
--- a/src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
+++ b/src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
@@ -90,6 +90,13 @@ pub struct BroadPhaseMultiSap {
     //       Another alternative would be to remove ColliderProxyId and
     //       just use a Coarena. But this seems like it could use too
     //       much memory.
+    #[cfg_attr(
+        feature = "serde-serialize",
+        serde(
+            serialize_with = "parry::utils::hashmap::serialize_hashmap_capacity",
+            deserialize_with = "parry::utils::hashmap::deserialize_hashmap_capacity"
+        )
+    )]
     colliders_proxy_ids: HashMap<ColliderHandle, BroadPhaseProxyIndex>,
     #[cfg_attr(feature = "serde-serialize", serde(skip))]
     region_pool: SAPRegionPool, // To avoid repeated allocations.
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -696,7 +696,6 @@ pub mod serde {
     ///
     /// Useful for [`std::collections::HashMap`] with a non-string key,
     /// which is unsupported by [`serde_json`].
-    #[cfg(feature = "serde-serialize")]
     pub fn deserialize_from_vec_tuple<
         'de,
         D: serde::Deserializer<'de>,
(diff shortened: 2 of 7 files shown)


src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
```diff
@@ -90,6 +90,13 @@ pub struct BroadPhaseMultiSap {
...
     //       just use a Coarena. But this seems like it could use too
     //       much memory.
+    #[cfg_attr(
+        feature = "serde-serialize",
+        serde(
+            serialize_with = "parry::utils::hashmap::serialize_hashmap_capacity",
+            deserialize_with = "parry::utils::hashmap::deserialize_hashmap_capacity"
+        )
+    )]
     colliders_proxy_ids: HashMap<ColliderHandle, BroadPhaseProxyIndex>,
```
sebcrozet, line 100: This doesn’t look correct. `serialize_hashmap_capacity` serializes the hashmap as a single value (its capacity) but doesn’t actually serialize its content. So deserialization will result in an emptied `collidiers_proxy_id` which is invalid.

I’m surprised this breaks string serialization though. Is it failing to serialize `ColliderHandle` or `BroadPhaseProxyIndex`?

sebcrozet (changes requested):

src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
Vrixyz: It was failing on the key ; I'll dig more

Vrixyz: I believe that's the root issue: https://github.com/serde-rs/json/issues/887. I'm surprised there is not a workaround provided in the issue :(

I imagine we could serialize to Vec, and deserialize to HashMap ? But I'm not sure how that would get implemented. And **where**, because we should keep the same hashmap serialization logic for bincode.

sebcrozet: Mmh, yeah, I’m surprised no alternative was suggested in the serde issue.

> I imagine we could serialize to Vec, and deserialize to HashMap ?

That sounds reasonable. Implementing this would be somewhat similar to `de/serialize_hashmap_capacity`. The serialization functions would look something like that:

```rust
/// Serializes only the capacity of a hash-map instead of its actual content.
#[cfg(feature = "serde-serialize")]
pub fn serialize_hashmap_as_vec<S: serde::Serializer, K, V, H: std::hash::BuildHasher>(
    map: &StdHashMap<K, V, H>,
    s: S,
) -> Result<S::Ok, S::Error> {
    let hashmap_as_vec = /* convert the hashmap to a vec */;
    hashmap_as_vec.serialize(s);
}

/// Creates a new hash-map with its capacity deserialized from `d`.
#[cfg(feature = "serde-serialize")]
pub fn deserialize_hashmap_capacity<
    'de,
    D: serde::Deserializer<'de>,
    K,
    V,
    H: std::hash::BuildHasher + Default,
>(
    d: D,
) -> Result<StdHashMap<K, V, H>, D::Error> {
    let hashmap_as_vec: Vec<(K, V)> = Deserialize::deserialize(deserializer)?;
    /* Convert the hashmap_as_vec to a hashmap */
}

```

> because we should keep the same hashmap serialization logic for bincode

It’s fine if we don’t keep the same logic for bincode.

src/utils.rs
```diff
@@ -667,3 +667,74 @@ pub fn smallest_abs_diff_between_angles<N: SimdRealCopy>(a: N, b: N) -> N {
...
+    ) -> Result<S::Ok, S::Error> {
+        let container: Vec<_> = target.into_iter().collect();
+        serde::Serialize::serialize(&container, s)
+    }
+
+    /// Deserializes from a `Vec<(K, V)>`.
+    ///
+    /// Useful for [`std::collections::HashMap`] with a non-string key,
+    /// which is unsupported by [`serde_json`].
+    #[cfg(feature = "serde-serialize")]
```
sebcrozet, line 699: Not needed since the `pub mod serde` is already feature-gated.

sebcrozet (approved):
Looking good, thanks!

src/utils.rs
Vrixyz: I was conflicted between gating the whole mod or each functions ; gating the whole mod might give users a bit of a pain for their "use" statements, but it's not a huge annoyance, and it feels cleaner to gate the whole mod

PR #679, opened 2024-07-12
Remove check for `target_arch = "asmjs"`

waywardmonkeys:
This is long obsolete and was removed from Rust itself in https://github.com/rust-lang/rust/pull/117338


Vrixyz (approved):

Vrixyz:
Thanks!

PR #680, opened 2024-07-12
Use `bitflags!` consistently.

waywardmonkeys:
This removes an `extern crate` and a `use` so that we always call it via `bitflags::bitflags!` everywhere.


Vrixyz (approved):

Vrixyz:
Thanks!

PR #681, opened 2024-07-12
Prefer explicit features for optional deps.

waywardmonkeys:
Implicit features are slated to be removed in a future version of Rust (2024 edition).

Fixing this exposed 2 instances where the wrong feature was being checked for `serde` vs `serde-serialize`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,7 +28,7 @@ pub extern crate parry3d_f64 as parry;
 
 pub extern crate crossbeam;
 pub extern crate nalgebra as na;
-#[cfg(feature = "serde")]
+#[cfg(feature = "serde-serialize")]
 #[macro_use]
 extern crate serde;
 extern crate num_traits as num;
--- a/crates/rapier2d-f64/Cargo.toml
+++ b/crates/rapier2d-f64/Cargo.toml
@@ -19,17 +19,17 @@ maintenance = { status = "actively-developed" }
 default = ["dim2", "f64"]
 dim2 = []
 f64 = []
-parallel = ["rayon"]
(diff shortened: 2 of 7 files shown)


waywardmonkeys:
With this, the list of features when using `cargo add` is closer to correct.

Before:

```
      Adding rapier3d v0.21.0 to dependencies
             Features:
             + dim3
             + f32
             - debug-disable-legitimate-fe-exceptions
             - debug-render
             - dev-remove-slow-accessors
             - enhanced-determinism
             - instant
             - parallel
             - profiler
             - rayon
             - serde
             - serde-serialize
             - simd-is-enabled
             - simd-nightly
             - simd-stable
             - vec_map
             - wasm-bindgen
```

After:

```
      Adding rapier3d (local) to dependencies
             Features:
             + dim3
             + f32
             - debug-disable-legitimate-fe-exceptions
             - debug-render
             - dev-remove-slow-accessors
             - enhanced-determinism
             - parallel
             - profiler
             - serde-serialize
             - simd-is-enabled
             - simd-nightly
             - simd-stable
             - wasm-bindgen
```

src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub extern crate parry3d_f64 as parry;
 
 pub extern crate crossbeam;
 pub extern crate nalgebra as na;
-#[cfg(feature = "serde")]
+#[cfg(feature = "serde-serialize")]
```
Vrixyz, line 31: :+1: ; I did see that recently, I'm glad you made a PR 🥇

Vrixyz (approved):

PR #682, opened 2024-07-12
Use workspace lints to control unexpected_cfgs.

waywardmonkeys:
By using `workspace.lints`, we can configure lints once for everything (and then inherit it into the various crates).

The lint configuration for `unexpected_cfgs` works in 1.80+ and warns otherwise.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,14 @@ members = ["crates/rapier2d", "crates/rapier2d-f64", "crates/rapier_testbed2d", 
     "crates/rapier3d", "crates/rapier3d-f64", "crates/rapier_testbed3d", "crates/rapier_testbed3d-f64", "examples3d", "examples3d-f64", "benchmarks3d", "crates/rapier3d-urdf", "crates/rapier3d-stl"]
 resolver = "2"
 
+[workspace.lints]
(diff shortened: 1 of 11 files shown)


waywardmonkeys:
I've marked this as a draft as there are unexpected cfgs still that need to be resolved:

* `log` in src_testbed
* `other-backends` isn't configured for `rapier_testbed3d-f64` while it is for the other 3 configurations.

This will generate a warning from `cargo` in Rust 1.79 and earlier (1.80 is due for release soon) and there are already 2 warnings from `cargo` that are more verbose than this one.

I think this replaces #638 from @urgau

To answer one of the concerns from @sebcrozet , advantages of this PR is that we get the benefit of the `unexpected_cfgs` lint again and we can start to use `workspace.lints` to centralize some other lint configurations in the future.

waywardmonkeys:
Now that #684 has landed, I've dealt with the `other-backends` question in here and this should be ready now.

I've confirmed that it builds without warnings on beta builds (warnings from `unexpected_cfgs`).

Vrixyz (approved):

Vrixyz:
Thanks!

waywardmonkeys:
You're welcome!

PR #684, opened 2024-07-12
testbed: Remove use of missing log feature

waywardmonkeys:
This code has been here but not used for a long time. There's no log feature (or dependency) within the testbed crates.


waywardmonkeys:
@sebcrozet any chance you could confirm that this is dead code and okay to remove?

PR #686, opened 2024-07-17
More Debug derives.

Vrixyz:


sebcrozet (approved):
Thanks!
This adds `Debug` on quite large structures. Have you noticed any reduction of compilation times after this change?

Vrixyz:
I didn't detect a meaningful compilation time change ; it seems to inflates the library size by 0.2MiB though:

image

(I also compared with full clean and results were also similar between master and this branch.)

PR #688, opened 2024-07-18
Fix `rustdoc::broken_intra_doc_links`

waywardmonkeys:
These show up when the `serde_serialize` feature is enabled.


Vrixyz:
good catch thanks

PR #690, opened 2024-07-18
clippy: Fix `clippy::field_reassign_with_default` lints

waywardmonkeys:


waywardmonkeys:
These are in a test target, so show up when using `--all-targets` with clippy.

Vrixyz:
Thanks, I imagine it would be a good idea to add that `--all-targets` to ci 💭

PR #691, opened 2024-07-18
Fix spelling in filename for `debug_excentric_boxes3`

waywardmonkeys:


Vrixyz (approved):

Vrixyz:
Thanks! I was surprised this didn't modify https://github.com/dimforge/rapier/blob/master/examples3d/all_examples3.rs ; but it looks like this particular example isn't covered by testbed examples. I'm curious if we can rework the testbed for an "auto-discovery" of examples... that wouldn't be trivial to make though.

PR #700, opened 2024-07-24
Fix spelling of `below` and `orthogonalization`.

waywardmonkeys:


Vrixyz:
Thanks! Would it make sense to add "Bellow" as a 'forbidden' word for the typo check ? it seems unlikely anything in our repo will "make the deep roaring sound characteristic of a bull."

Also would it make sense to make an upstream issue to typos for `ortogonalization` ?

waywardmonkeys:
Up to you!

Submitting upstream is a matter of commenting on a bug, I think.

Vrixyz:
I Filed https://github.com/crate-ci/typos/issues/1051#issuecomment-2252649363 as a follow up

Issue #467, opened 2023-03-18
Add optional bevy feature for reflection


janhohenheim:
Currently, `bevy_rapier` can (de facto) not be used in tandem with bevy dynamic scenes, since colliders cannot be serialized due to not being `Reflect`. This means that save systems that would like to use the built-in bevy scene format need to clutter the graph with marker traits for which colliders were spawned how, which defeats the purpose of the dynamic scene as "just dump the entire world into a file and don't worry about it". Additionally, colliders cannot be inspected at runtime in editors, which is a pity if generating them dynamically.

I propose adding an optional `bevy_reflect` feature to rapier that only derives `Reflect` and `FromReflect` for the types that aren't in `bevy_rapier` and then add them to the type registry through `bevy_rapier`.

If I get your okay @sebcrozet, I'll do this myself.

sebcrozet:
Could `Reflect` just be implemented directly (manually) on the wrapper types defined by `bevy_rapier` instead of having to be propagated down to `rapier` (and, most likely, `parry`)?

sebcrozet:
There is no plan to add a `Reflect` derive to `rapier` since that would also require such feature on `parry`, and `nalgebra`. We should  have `Reflect` implemented manually in `bevy_rapier` types instead.

janhohenheim:
Last time I checked (1 year ago), this was impossible to do properly as the underlying types inside the wrappers did not derive `Reflect`. There is a PR in the works in Bevy that allows deriving `Reflect` on foreign types by redefining them and then deriving `Reflect` on these dummy types. You can do that, just be aware that this means duplicating many `parry` types for `bevy_rapier`.

sebcrozet:
Thank you for the info! We’ll keep an eye on this. Hopefully we can work something out at the `bevy_rapier` level eventually.

PR #701, opened 2024-07-24
Fix character max/min slope

Vrixyz:
This PR has a bit too many different changes, but bare with me:
- Initially, I was investigating on https://github.com/dimforge/bevy_rapier/pull/532.

This PR approaches:

## character controller `max_slope` bug

- While testing rapier, I thought rapier did not have this bug, but looking more carefully, the setup for the slope we can't climb is slightly incorrect and we should jump on it to check that we're blocked.

<details><summary>screenshot testbed</summary>
<p>

We probably want the bigger slope to be reachable without jumping.

image

</p>
</details> 

⚠️ So the slope bug seems to originate from rapier, and not bevy / javascript integration.

## Testbed character controller parameters

- character controller has some parameters, which are note exposed in the testbed, making it difficult to quickly test a behaviour.
  - to be noted it would be more maintainable and robust with a reflect feature for rapier, this argument wasn't brought on https://github.com/dimforge/rapier/issues/467.
<details><summary>Screenshot controller exposed</summary>
<p>

WIP 🚧, it lacks some options and a title group, same controls could be added to the vehicle controller.

image

</p>
</details> 

## Testbed update to bevy 0.14

- a more focused PR on this is https://github.com/dimforge/rapier/pull/723


examples3d/character_controller3.rs
```diff
@@ -94,19 +94,15 @@ pub fn init_world(testbed: &mut Testbed) {
...
-    let collider = ColliderBuilder::cuboid(
-        slope_size * scale,
-        ground_height * scale,
-        slope_size * scale,
-    )
-    .translation(vector![ground_size + slope_size, -ground_height + 0.4, 0.0] * scale)
-    .rotation(Vector::z() * slope_angle);
+    let collider = ColliderBuilder::cuboid(slope_size, ground_height, slope_size)
+        .translation(vector![0.1 + slope_size, -ground_height + 0.4, 0.0])
+        .rotation(Vector::z() * slope_angle);
```
Vrixyz, line 99: scale to add back ; not sure it's still usable though 🤔

src/control/character_controller.rs
```diff
@@ -546,28 +554,29 @@ impl KinematicCharacterController {
...
+        // An object is slipping if the tangential movement induced by its vertical movement points downward.
         let slipping = self.up.dot(&decomp.vertical_tangent) < 0.0;
 
-        // An object is trying to climb if its indirect vertical motion points upward.
-        let climbing_intent = self.up.dot(&input_decomp.vertical_tangent) > 0.0;
-        let climbing = self.up.dot(&decomp.vertical_tangent) > 0.0;
+        // An object is trying to climb if its vertical (according to self.up) input motion points upward.
+        let climbing_intent = self.up.dot(&_vertical_input) > 0.0;
+        // An object is climbing if the tangential movement induced by its vertical movement points upward.
+        let climbing = self.up.dot(&decomp.vertical_tangent) > 0.001;
```
Vrixyz, line 566: maybe a comment to justify that would be useful, and use the explicit epsilon ?

src/control/character_controller.rs
```diff
@@ -546,28 +554,29 @@ impl KinematicCharacterController {
...
 
-        let allowed_movement = if hit.is_wall && climbing && !climbing_intent {
+        let allowed_movement = if hit.is_wall && (climbing && !climbing_intent) {
```
Vrixyz, line 568:
```suggestion
        let allowed_movement = if hit.is_wall && climbing && !climbing_intent {
```

src/control/character_controller.rs
```diff
@@ -546,28 +554,29 @@ impl KinematicCharacterController {
...
             // Can’t climb the slope, remove the vertical tangent motion induced by the forward motion.
             decomp.horizontal_tangent + decomp.normal_part
```
Vrixyz, line 570: when blocking climb, we don't want to apply the nudge factor, right ?

src/control/character_controller.rs
```diff
@@ -904,3 +913,164 @@ fn subtract_hit(translation: Vector<Real>, hit: &ShapeCastHit) -> Vector<Real> {
...
     translation + *hit.normal1 * surface_correction
 }
+
+#[cfg(test)]
+mod test {
+    use std::default;
+
+    use crate::{control::KinematicCharacterController, prelude::*};
+
+    #[cfg(all(feature = "dim3", feature = "f32"))]
```
Vrixyz, line 923: I gated it behind features because it's already a big test, I fear adding support for multiple dimensions and precision would make it even less approachable

src_testbed/ui.rs
```diff
@@ -240,7 +241,57 @@ pub fn update_ui(
...
+    /// This is a small distance applied to the movement toward the contact normals of shapes hit
+    /// by the character controller. This helps shape-casting not getting stuck in an always-penetrating
+    /// state during the sliding calculation.
+    ///
+    /// This value should remain fairly small since it can introduce artificial "bumps" when sliding
+    /// along a flat surface.
+    pub normal_nudge_factor: Real,
+*/
+        }
+        ui.separator();
```
Vrixyz, line 294: this was copied to easily know which parameters I had to "inspect", I didn't go all the way.

src/control/character_controller.rs
```diff
@@ -269,7 +271,13 @@ impl KinematicCharacterController {
...
                 ShapeCastOptions {
                     target_distance: offset,
-                    stop_at_penetration: false,
+                    stop_at_penetration:
+                    // We want to know the collision we're currently in,
+                    // otherwise we risk ignoring useful collision information (walls / ground...)
+                    iter_index == 0
+                    // If we stop at penetration while not allowing sliding alongside shapes, 
+                    // it will be impossible to get away of a collision.
+                    && self.slide,
```
Vrixyz, line 280: part 1 of the fix

src/control/character_controller.rs
```diff
@@ -546,28 +554,29 @@ impl KinematicCharacterController {
...
+        let slipping_intent = self.up.dot(&_vertical_input) < 0.0;
+
+        // An object is slipping if the tangential movement induced by its vertical movement points downward.
         let slipping = self.up.dot(&decomp.vertical_tangent) < 0.0;
 
-        // An object is trying to climb if its indirect vertical motion points upward.
-        let climbing_intent = self.up.dot(&input_decomp.vertical_tangent) > 0.0;
-        let climbing = self.up.dot(&decomp.vertical_tangent) > 0.0;
+        // An object is trying to climb if its vertical (according to self.up) input motion points upward.
+        let climbing_intent = self.up.dot(&_vertical_input) > 0.0;
```
Vrixyz, line 564: part 2 of the fix: using input to infer "user intent", rather than movement tangent.

This leads to unused `horiz_input_decomp` and `input_decomp`.

src/control/character_controller.rs
Vrixyz: Actually, this fixes ground detection but only by side effect: the condition with `// No interference along the path.` contains a break, and `grounded` is computed just after. I think we actually want to computed the grounded status after the loop.

src/control/character_controller.rs
```diff
@@ -321,23 +322,23 @@ impl KinematicCharacterController {
...
+            bodies,
+            colliders,
+            queries,
+            character_shape,
+            &(Translation::from(result.translation) * character_pos),
+            &dims,
+            filter,
+            Some(&mut kinematic_friction_translation),
+            Some(&mut translation_remaining),
+        );
```
Vrixyz, line 341: part 1 of the fix (the break from line 322 (in block `// No interference along the path.`) would bypass the ground detection, resulting in `not grounded`. That's happening because we're not checking for current colliding shapes. when doing the cast.



reworked from https://github.com/dimforge/rapier/pull/701/files#r1691673551

src/control/character_controller.rs
sebcrozet: Mmh, I think I tried that at some point and this had a side-effect of having the user climb some slopes it shouldn’t. It would be worth moving around in the character-controller scene.

This also makes me think that perhaps we should make a test scene where we have many character-controller (in isolation) that try to walk on various surface configurations. That would help spot regressions more easily without having to move the character manually.

src/control/character_controller.rs
Vrixyz: Then we still need to avoid climbing as a side effect 🤔 I think.

Not sure on the correct math to do that 😅 ; mayyybe we should cross that horizontal tangeant with self.up

src/control/character_controller.rs
Vrixyz: a scene with multiple character controller is an interesting idea ; to be noted I added an automated test for character controller.

src/control/character_controller.rs
Vrixyz: after testing, removing the nudge factor doesn't seem to fix the "can climb forbidden slopes", so I'm reverting that change. To be clear, the current behaviour is still better than what's on master.

src/control/character_controller.rs
```diff
@@ -163,12 +163,13 @@ impl Default for KinematicCharacterController {
             max_slope_climb_angle: Real::frac_pi_4(),
             min_slope_slide_angle: Real::frac_pi_4(),
             snap_to_ground: Some(CharacterLength::Relative(0.2)),
-            normal_nudge_factor: 1.0e-4,
+            normal_nudge_factor: 1.0e-5,
```
Vrixyz, line 166: This was messing up the grounded status on the test scene ; if that's a bad default, more explaination and/or a test case for this default should be added.

src/control/character_controller.rs
Vrixyz: after more testing with technique from https://github.com/dimforge/rapier/pull/705 ; this doesn't seem to improve significantly the behaviour, so I'll leave it out of this PR and focus on the slope. This PR builds a useful test to plus such verification into it though.

src/control/character_controller.rs
```diff
@@ -542,17 +541,18 @@ impl KinematicCharacterController {
...
         let slipping_intent = self.up.dot(&horiz_input_decomp.vertical_tangent) < 0.0;
+
+        // An object is slipping if its vertical movement points downward.
         let slipping = self.up.dot(&decomp.vertical_tangent) < 0.0;
 
-        // An object is trying to climb if its indirect vertical motion points upward.
-        let climbing_intent = self.up.dot(&input_decomp.vertical_tangent) > 0.0;
+        // An object is trying to climb if its vertical input motion points upward.
+        let climbing_intent = self.up.dot(&_vertical_input) > 0.0;
+        // An object is climbing if the tangential movement induced by its vertical movement points upward.
```
Vrixyz, line 555: in my understanding, "climbing intent" refers to an explicit movement going upward of the slope, that's why I'm testing only the input, and not the decomposition according to the hit.


the naming is confusing because slipping_intent doesn't refer exactly to the same thing.

src/control/character_controller.rs
```diff
@@ -904,3 +904,170 @@ fn subtract_hit(translation: Vector<Real>, hit: &ShapeCastHit) -> Vector<Real> {
...
+        let slope_size = 2.0;
+        let collider = ColliderBuilder::cuboid(slope_size, ground_height, slope_size)
+            .translation(vector![0.1 + slope_size, -ground_height + 0.4, 0.0])
+            .rotation(Vector::z() * slope_angle);
+        colliders.insert(collider);
+
+        /*
+         * Create a slope we can’t climb.
+         */
+        let impossible_slope_angle = 0.6;
```
Vrixyz, line 953: on master, this impossible angle can be climb in this test. That's what https://github.com/dimforge/rapier/pull/701/files#r1697044257 fixes.

sebcrozet (approved):

PR #706, opened 2024-08-02
Use a reserved "Default" layer for memberships in `InteractionGroups`

Aceeri:
Currently the default is to being a member of all groups with a filter/interaction with everything.

I believe this is a bit backwards and makes special interactions between groups harder to understand/deal with without flipping all of your logic to be a "non-membership" or by modifying the interaction groups of every object in your simulation.

This is more in line with what Box2d (one membership, all filter) and Godot (one membership, one filter) do currently, ~however there are counterexamples like Unity (all membership, all filter) and the current default.~ Unity is actually one membership/all filter as well since they do not allow multiple groups to be selected, however they allow configuring the default interactions/filters.

There is also an argument of all filter vs one filter as the default, but I do not believe it matters as much as the membership.

### DIFF
--- a/src/geometry/interaction_groups.rs
+++ b/src/geometry/interaction_groups.rs
@@ -72,7 +72,7 @@ impl InteractionGroups {
 impl Default for InteractionGroups {
     fn default() -> Self {
         Self {
-            memberships: Group::DEFAULT,
+            memberships: Group::GROUP_1,
             filter: Group::ALL,
         }
     }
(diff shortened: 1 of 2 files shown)


src/geometry/interaction_groups.rs
```diff
@@ -80,70 +83,70 @@ bitflags::bitflags! {
     #[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
     #[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
     pub struct Group: u32 {
+        /// The default group.
+        const DEFAULT = 1 << 0;
         /// The group n°1.
-        const GROUP_1 = 1 << 0;
```
Vrixyz, line 84: Not sure I agree on shifting those groups though:
- it makes a tedious migrations for users
- javascript bindings will have to adapt (not hard, but tedious/error prone again)

the naming becomes a bit inconsistent, I would prefer `DEFAULT` and `GROUP_1` being the same, or straight up removing DEFAULT, and add that's the default as a comment.

Aceeri: Ya I agree

Vrixyz:
Thanks!

PR #707, opened 2024-08-02
Fix CI

Vrixyz:
These new warnings come from rust 1.80.

I'd like to take this opportunity to fix unused `NodeReferences`, `Externals` and `Neighbors` ; I imagine they should be exposed, and not straight up deleted.

### DIFF
--- a/src/data/mod.rs
+++ b/src/data/mod.rs
@@ -2,7 +2,6 @@
 
 pub use self::arena::{Arena, Index};
 pub use self::coarena::Coarena;
-pub use graph::{Externals, Neighbors, NodeReferences};
 
 pub mod arena;
 mod coarena;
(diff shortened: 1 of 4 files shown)


src/data/mod.rs
```diff
@@ -2,6 +2,7 @@
 
 pub use self::arena::{Arena, Index};
 pub use self::coarena::Coarena;
+pub use graph::{Externals, Neighbors, NodeReferences};
```
Vrixyz, line 5: Not entirely sure about this. it seems practical.

Another way to expose it would be through the graph module, but the `pub(crate) mod graph;` means that we'd have to change all public fields within `graph` to `pub(crate)`

sebcrozet: Let’s remove the dead code. We can revive it from the git history if we really need it in the future.

sebcrozet (approved):

sebcrozet:
Thanks!

PR #708, opened 2024-08-04
Fix shape modification not updating graphics in testbed

agarret7:
This PR is for a fix for the `GraphicsManager::remove_collider_nodes` function in the testbed. The problem was that the entity was despawned but not removed from `GraphicsManager::b2sn`, creating an invalid state. See [this thread](https://discord.com/channels/507548572338880513/1246940826332958801) for the discussion.

`TestbedApp::add_callback` still requires a manual refresh with `gfx.add_collider` if the user wants to see the 3D entity update correctly.

```rust
if let Some(gfx) = &mut gfx {
    gfx.remove_collider(ball_coll_handle, &physics.colliders);
    gfx.add_collider(ball_coll_handle, &physics.colliders);
}
```

`debug_shape_modification3` before:

before

`debug_shape_modification3` after:

after

### DIFF
--- a/src_testbed/graphics.rs
+++ b/src_testbed/graphics.rs
@@ -82,14 +82,18 @@ impl GraphicsManager {
         collider: ColliderHandle,
     ) {
         let body = body.unwrap_or(RigidBodyHandle::invalid());
-        if let Some(sns) = self.b2sn.get_mut(&body) {
-            for sn in sns.iter_mut() {
+        if let Some(sns) = self.b2sn.remove(&body) {
+            let mut new_sn = vec![];
+            for sn in sns.into_iter() {
                 if let Some(sn_c) = sn.collider {
                     if sn_c == collider {
                         commands.entity(sn.entity).despawn();
+                    } else {
+                        new_sn.push(sn);
                     }
                 }
             }
+            self.b2sn.insert(body, new_sn);
         }
     }
 
--- a/src_testbed/graphics.rs
+++ b/src_testbed/graphics.rs
@@ -83,17 +83,17 @@ impl GraphicsManager {
     ) {
         let body = body.unwrap_or(RigidBodyHandle::invalid());
         if let Some(sns) = self.b2sn.remove(&body) {
-            let mut new_sn = vec![];
+            let mut new_sns = vec![];
             for sn in sns.into_iter() {
                 if let Some(sn_c) = sn.collider {
                     if sn_c == collider {
                         commands.entity(sn.entity).despawn();
                     } else {
-                        new_sn.push(sn);
+                        new_sns.push(sn);
                     }
                 }
             }
-            self.b2sn.insert(body, new_sn);
+            self.b2sn.insert(body, new_sns);
         }
     }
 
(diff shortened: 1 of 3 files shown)


Vrixyz (approved):

src_testbed/graphics.rs
```diff
@@ -78,14 +78,18 @@ impl GraphicsManager {
...
+            for sn in sns.into_iter() {
                 if let Some(sn_c) = sn.collider {
                     if sn_c == collider {
                         commands.entity(sn.entity).despawn();
+                    } else {
+                        new_sns.push(sn);
                     }
                 }
             }
+            self.b2sn.insert(body, new_sns);
```
Vrixyz, line 92:
```suggestion
        if let Some(mut sns) = self.b2sn.remove(&body) {
            sns = sns
                .into_iter()
                .filter(|sn| {
                    let Some(sn_c) = sn.collider else {
                        return false;
                    };
                    if sn_c == collider {
                        commands.entity(sn.entity).despawn();
                        return false;
                    } else {
                        return true;
                    }
                })
                .collect();
            self.b2sn.insert(body, sns);
        }
```

Small nitpick, I did that just to help me understand the change, not sure if which is better, don´t worry about it I'll merge one or the other after discussion with seb.

Vrixyz:
For the record, I think the best solution is to monitor changes over the shapes, and then modify the mesh handle:

In practice, running these snippets of code:

- recompute the mesh:

https://github.com/dimforge/rapier/blob/eba44e71f8cee93a835bdc690dd6ea7d8053909a/src_testbed/objects/node.rs#L82-L85

- Then reapply the mesh handle on the entity.

But as we're in testbed, I think this solution is good enough :)

agarret7:
I had thought that it might be good to factor these sorts of updates into the testbed loop, but wasn't sure how to track incremental computation of the state. Especially since I noticed you have a vector of flags already tracking collider updates in `ColliderSet::changes` private to the `rapier3d` crate and that `collider.set_shape` uses this internally:

https://github.com/dimforge/rapier/blob/eba44e71f8cee93a835bdc690dd6ea7d8053909a/src/geometry/collider.rs#L445-L449

It seems like monitoring those changes from the testbed requires a new function like `pub fn get_changes() -> &ColliderChanges`? From what I could tell the `ColliderChanges` otherwise cannot be seen outside the main crate and thus wasn't sure how you might select meshes to be recomputed.

src_testbed/graphics.rs
sebcrozet: This can be simplified further with `Vec::retain`:

```rust
        if let Some(sns) = self.b2sn.get_mut(&body) {
            sns.retain(|sn| {
                if let Some(sn_c) = sn.collider {
                    if sn_c == collider {
                        commands.entity(sn.entity).despawn();
                        return false;
                    }
                }

                true
            });
        }
```
this also fixes the case where we the scene node doesn’t have a collider attached, in which case we don’t want this method to remove it from the list.

sebcrozet: (I applied the change.)

sebcrozet:
Thanks!

PR #705, opened 2024-07-30
Color visualization for grounded character controller

Vrixyz:
It's currently difficult to assess if character controller ground detection works correctly, this adds a visual helper (the capsule turns red when not grounded) to detect it more quickly in the examples.

https://github.com/user-attachments/assets/f0a5b817-4563-46db-a6fe-16d5433d91b3

<details><summary>ground detection analysis</summary>
<p>

when calling `colliders_with_aabb_intersecting_aabb` from within `detect_grounded_status_and_apply_friction`
, it sometimes doesn't get any manifests, resulting in `grounded = false`.

- Honestly, a raycast down is probably less work than fiddling with nudge factor, numerical errors, shape implementations and whatnot 🤔 

- Or maybe just exposing or bumping up the predict ground distance around https://github.com/dimforge/rapier/blob/661c33f64c5a3392f5abb8a200151be8580ec9e5/src/control/character_controller.rs#L400

- I tested it against https://github.com/dimforge/rapier/pull/481, and still have not grounded frames.

</p>
</details>

PR #711, opened 2024-08-05
fix character controller is_grounded 

Vrixyz:
- in https://github.com/dimforge/rapier/pull/446 ; there had been a lot of experimentation on that offset. this `1.1` has been kept since the creation of the character controller.
- I think that 1.1 is not enough, I'm not sure exactly why there's number approximation this high, but manual testing results in better behaviour (no red controller from https://github.com/dimforge/rapier/pull/705), and automatic testing still passes.

## Some numbers

offset is `CharacterLength::Relative(0.01)`, then the height of the `CharacterController` in the exampple is `0.3`, resulting in an exxaggerated  offset (for aabb computation) of  `0.0033`. The nudge factor being `0.0001` shouldn´t impact that, but we're still in small numbers game. I think overshooting a bit won´t hurt much, especially if it fixes `is_grounded` reliability.

## TODO?
I think I can make a test about ground prediction (make a character walking on a trimesh for some time, verify it's always grounded), but I fear this test might not always catch such bug.

## Alternatives

We can consider exposing this offset multiplier, or the whole AABB used.

### DIFF
--- a/src/control/character_controller.rs
+++ b/src/control/character_controller.rs
@@ -397,7 +397,7 @@ impl KinematicCharacterController {
     }
 
     fn predict_ground(&self, up_extends: Real) -> Real {
-        self.offset.eval(up_extends) * 1.1
+        self.offset.eval(up_extends) * 1.2
     }
 
     fn detect_grounded_status_and_apply_friction(


sebcrozet (approved):
It would be interesting to know why `1.1` didn’t work here, but let’s merge this as a stopgap solution. Thanks!

Vrixyz:
> It would be interesting to know why `1.1` didn’t work here, but let’s merge this as a stopgap solution. Thanks!

@sebcrozet My behaviour analysis, with a capsule moving on a flat ground (character controller example)

It's frustrating that I can´t find why the capsule ends up being too high, but I'll give up on further analysis for now...

image

https://docs.google.com/presentation/d/1Uc9HEgNH0H3Mf9zvaiuNFCdopPnl50QVsyTkjxdK4po/edit?usp=sharing

PR #715, opened 2024-08-12
Fix character controller ground detection attempt 3

Vrixyz:
Following details from https://github.com/dimforge/rapier/pull/711#issuecomment-2283694087

- always call detect_ground (which was not done if no collision were initially detected, or if we were not moving).
- flat value for `predict_ground` function

The tests were not passing though, a ground of 600 makes the controller not grounded at
```
  ┌             ┐
  │   502.99017 │
  │ -0.19247428 │
  │   100.54156 │
  └             ┘
```

By adding more height to the ground, I could make the test pass.

image

:warning: This PR poses a problem of detecting ground when too near of a ground, because loosened shape intersect with it.

<details><summary>video illustration</summary>
<p>

Unfortunately we can´t see in the video, but during a frame the charactercontroller is detected as grounded when flying exactly on the edge of the "cliff":

[Screencast from 09-09-2024 04:21:37 PM.webm](https://github.com/user-attachments/assets/0a278503-9160-4804-9433-d6cbb8042d8f)

</p>
</details> 

It's not the case before because we were not computing the ground state when there was no collision on the initial `cast_shape`.


sebcrozet (approved):
Looking good, thanks.
It’s interesting that you sometimes get `minY ~= 0.02`. I’m wondering if the offset gets counted twice somewhere (because 0.02 ~= 0.009 * 2 + 0.004).

PR #726, opened 2024-08-30
Improve capacity handling for `ColliderSet`, `RigidBodySet`.

waywardmonkeys:
These allow an application to reduce the cost of reallocation when they know that a large number of colliders or rigid bodies will be created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - The region key has been replaced by an i64 in the f64 version of rapier, increasing the range before panics occur.
 - Fix `BroadphaseMultiSap` not being able to serialize correctly with serde_json.
 
+### Added
+
+- `RigidBodySet` and `ColliderSet` have a new constructor `with_capacity`.
+
 ### Modified
 
 - `InteractionGroups` default value for `memberships` is now `GROUP_1` (#706)
--- a/src/dynamics/rigid_body_set.rs
+++ b/src/dynamics/rigid_body_set.rs
@@ -43,6 +43,14 @@ impl RigidBodySet {
         }
     }
 
+    /// Create a new set of rigid bodies, with an initial capacity.
+    pub fn with_capacity(capacity: usize) -> Self {
+        RigidBodySet {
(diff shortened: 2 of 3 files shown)


waywardmonkeys:
We may at times create a scene with a large number of colliders and prefer that doing so be fast. (In our case, we're not running physics or rigid bodies. We're also able to disable the colliders as we don't need the full collision detection, just answers to queries.)

For a silly test case of 8,000,000 (200x200x200) things (using `many_static.rs` as a starter), this reduced the time for populating the sets from 1.3s to 1.0s.

There's a number of questions that could be considered here beyond just these changes:

* Should other sets get this sort of thing?
* Should there be a way to set the capacity for the `removed_colliders` on `ColliderSet`?
* Should we give access to methods for `capacity`, `reserve`, `shrink_to`?

waywardmonkeys:
@Vrixyz @sebcrozet Just checking in on this...

Vrixyz (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 - The region key has been replaced by an i64 in the f64 version of rapier, increasing the range before panics occur.
 - Fix `BroadphaseMultiSap` not being able to serialize correctly with serde_json.
 
+### Added
+
+- `RigidBodySet` and `ColliderSet` have a new constructor `with_capacity`.
```
waywardmonkeys, line 10: Should this point to this PR?

Vrixyz: Ideally yes, but it's rare to have the PR link, I'm adding it arbitrarily on bigger/breaking/difficult PRs

Vrixyz:
I'll merge this as-is ; as for your questions, 


>    Should other sets get this sort of thing?

A good idea.

>    Should there be a way to set the capacity for the removed_colliders on ColliderSet?
If there's a real use case for it, I'd approve it, but probably pass in the exact values of each properties, so each properties can be tweaked ?

>    Should we give access to methods for capacity, reserve, shrink_to?
Probably a good idea, but I'd prefer a real use case too.

Issue #720, opened 2024-08-20
urdf3.rs doesn't work cause of version discrepancies


Guelakais:
by running your urdf3.rs example locally, I got the following error:
```sh
error[E0308]: arguments to this method are incorrect
   --> src/urdf.rs:33:10
    |
33  |         .insert_using_impulse_joints(&mut bodies, &mut colliders, &mut impulse_joints);
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^ -----------  --------------  ------------------- expected `ImpulseJointSet`, found `rapier3d::dynamics::ImpulseJointSet`
    |                                      |            |
    |                                      |            expected `rapier3d::geometry::collider_set::ColliderSet`, found `rapier3d::geometry::ColliderSet`
    |                                      expected `rapier3d::dynamics::rigid_body_set::RigidBodySet`, found `rapier3d::dynamics::RigidBodySet`
    |
    = note: `rapier3d::dynamics::RigidBodySet` and `rapier3d::dynamics::rigid_body_set::RigidBodySet` have similar names, but are actually distinct types
note: `rapier3d::dynamics::RigidBodySet` is defined in crate `rapier3d`
   --> /home/deck/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rapier3d-0.22.0/src/dynamics/rigid_body_set.rs:28:1
    |
28  | pub struct RigidBodySet {
    | ^^^^^^^^^^^^^^^^^^^^^^^
note: `rapier3d::dynamics::rigid_body_set::RigidBodySet` is defined in crate `rapier3d`
   --> /home/deck/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rapier3d-0.21.0/src/dynamics/rigid_body_set.rs:28:1
    |
28  | pub struct RigidBodySet {
    | ^^^^^^^^^^^^^^^^^^^^^^^
    = note: perhaps two different versions of crate `rapier3d` are being used?
    = note: `rapier3d::geometry::ColliderSet` and `rapier3d::geometry::collider_set::ColliderSet` have similar names, but are actually distinct types
note: `rapier3d::geometry::ColliderSet` is defined in crate `rapier3d`
   --> /home/deck/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rapier3d-0.22.0/src/geometry/collider_set.rs:10:1
    |
10  | pub struct ColliderSet {
      | ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #727, opened 2024-09-02
chore: add publish script for urdf and stl + unify all releases by de…

Vrixyz:
- related to #720.
- Adds a script to publish urdf and stl
- default publish script publishes all crates, as they all depend on rapier, it's our most common usage.
- default behaviour for the root publish script is to run a dry-run, for being careful about it, and allow easier testing.
- :warning: This needs https://github.com/dimforge/rapier/pull/728 in order to build (I added a CI step to validate it fails, which will get green after that other PR)


sebcrozet:
Thanks!

PR #736, opened 2024-09-21
Fix Nans from RevoluteJoint::angle()

### LINKED ISSUE #735: `RevoluteJoint::angle()` occasionally incorrectly returns `NaN`
As far as I can see in my use-case, this is due to numeric imprecision which causes the `i` value of the `ang_err` quaternion to be marginally outside the range -1.0 to 1.0; which in turn causes `i.asin()` to return `NaN`.

A simple fix is to clamp the value passed to `i.asin()` to force it to the range -1.0 to 1.0

chrisdunelm:
Fixes #735


sebcrozet:
Thanks!

PR #716, opened 2024-08-12
ImpulseJointSet::get_mut option to wake up connected bodies

Vrixyz:
follow up to https://github.com/dimforge/rapier/issues/692

PR #739, opened 2024-09-22
fix compilation for feature enhanced-determinism

Vrixyz:
- Fix #739

- The regression was introduced by https://github.com/dimforge/rapier/pull/716

### DIFF
--- a/src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
+++ b/src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
@@ -232,7 +232,7 @@ impl BroadPhaseMultiSap {
          */
         for removed in removed_colliders {
             #[cfg(feature = "enhanced-determinism")]
-            let proxy_id = self.colliders_proxy_ids.shift_remove(removed);
+            let proxy_id = self.colliders_proxy_ids.swap_remove(removed);
             #[cfg(not(feature = "enhanced-determinism"))]
             let proxy_id = self.colliders_proxy_ids.remove(removed);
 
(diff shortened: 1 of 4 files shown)


Ughuuu:
I solved it similar to this impl too. It's annoying but it's good enough. Adding it to CI will help in future too. Best way to solve would be to make sure the API's match in parry, but it's not a huge problem (would be a huge problem if we had 100's of places of incompatibility).

Vrixyz:
Best way would be to have indexmap's API match HashMap right ? 

I'll probably merge the fix without CI and think about CI next.

But yeah in the meantime I was considering:
- add a trait for implementing those similar functions
- 'cfg if' in every lines
- have another CI step without blocking warnings 😅

Ughuuu:
Maybe best way would be that, but for now best would be to quickly solve this indeed so we don't break anyone who uses master.
I would also say the API should match standard API, but it's harder to do since we use another package that implements that, and we don't have much control over the API contracts.

src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs
```diff
@@ -231,7 +231,12 @@ impl BroadPhaseMultiSap {
          * Actually remove the colliders proxies.
          */
         for removed in removed_colliders {
-            if let Some(proxy_id) = self.colliders_proxy_ids.remove(removed) {
+            #[cfg(feature = "enhanced-determinism")]
+            let proxy_id = self.colliders_proxy_ids.shift_remove(removed);
+            #[cfg(not(feature = "enhanced-determinism"))]
+            let proxy_id = self.colliders_proxy_ids.remove(removed);
```
Vrixyz, line 237: These are modified because of a deprecated warning:

```
warning: use of deprecated method `indexmap::map::IndexMap::<K, V, S>::remove`: `remove` disrupts the map order -- use `swap_remove` or `shift_remove` for explicit behavior.
   --> crates/rapier3d/../../src/geometry/broad_phase_multi_sap/broad_phase_multi_sap.rs:234:62
    |
234 |             if let Some(proxy_id) = self.colliders_proxy_ids.remove(removed) {
    |                                                              ^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default
```

sebcrozet (approved):
👍

Issue #622, opened 2024-04-19
Feature: Add support for custom friction and bounce CombineRules


Ughuuu:
I have a use case where in Godot they use friction and bounce like so:
```
real_t combine_bounce(GodotBody3D *A, GodotBody3D *B) {
    return CLAMP(A->get_bounce() + B->get_bounce(), 0, 1);
}

real_t combine_friction(GodotBody3D *A, GodotBody3D *B) {
    return ABS(MIN(A->get_friction(), B->get_friction()));
}
```

So bounce they sum them, and then either take 0 or 1(bounce can be negative on some objects)
Friction they take the min and then take abs(agian, friction can be negative, and in such a case that is the value you always take). Weird case, but am thinking how could I do this in rapier?
Maybe if I could override the combine function from CoefficientCombineRule, I could make it work.

Ughuuu:
Any updates on this? I would be willing to implement it if there was some guidance as to what is wanted for this (eg. a callback, more enum types for different friction/bounce actions? etc)

Ughuuu:
Ended up implementing it in a fork of rapier, since it's probably not a feature that's much needed.

PR #170, opened 2021-04-09
Add InteractionGroups::test_or, increase to 64 bits

Demindiro:
The `InteractionTestMode` is analogous to Bullet's `collisionFilterMode` (see their [PyBullet quickstart guide][pyb], page 39).

Increasing the bits from 32 to 64 is useful when porting Rapier to other engines that already use 64 bit filters (e.g. Godot).

[pyb]: https://raw.githubusercontent.com/bulletphysics/bullet3/master/docs/pybullet_quickstartguide.pdf

PR #741, opened 2024-09-22
Implement interaction groups test mode and cofficient combine rule

Ughuuu:
- Continues https://github.com/dimforge/rapier/pull/170
- Fixes https://github.com/dimforge/rapier/issues/622


Ughuuu:
Ok, added this mainly to stop using my fork and to try to use rapier default in godot-rapier. This way I hope to no longer need to maintain that and use this repo as it is. (if this gets merged)

src/geometry/interaction_groups.rs
```diff
@@ -23,25 +30,39 @@ pub struct InteractionGroups {
...
 impl InteractionGroups {
     /// Initializes with the given interaction groups and interaction mask.
-    pub const fn new(memberships: Group, filter: Group) -> Self {
+    pub const fn new(memberships: Group, filter: Group, test_mode: InteractionTestMode) -> Self {
         Self {
             memberships,
             filter,
+            test_mode,
         }
     }
```
Vrixyz, line 56: To avoid some API breakage, we could add a `new_with_testmode` constructor, but it's not a very difficult migration so I'm fine with this.

CHANGELOG.md
```diff
@@ -1,12 +1,24 @@
...
 
+- The way contacts are represented changed. Refer to the documentation
+  of `parry::query::ContactManifold`, `parry::query::TrackedContact`
+  and `rapier::geometry::ContactManifoldData` and `rapier::geometry::ContactData` for details.
+
+### Breaking changes
+
+- `InteractionGroups` struct now contains `InteractionTestMode`.
+- `InteractionGroups` constructor now requires an `InteractionTestMode` parameter. If you want same behaviour as before, use `InteractionTestMode::AND` (eg. `InteractionGroups::new(Group::GROUP_1, Group::GROUP_1, InteractionTestMode::AND)`)
+- `CoefficientCombineRule::Min` - now makes sure it uses a non zero value as result by using `coeff1.min(coeff2).abs()`
```
Vrixyz, line 11: Thanks for filling the changelog ❤️. I'm happy with that description!

Previous entries of changelog don't have a particular "braking change" section and most end up in the "Modified" section, but that's details which can be edited on merge 👌 .

As it's not a trivial change, I'll probably want to add a link to that PR for additional context for the reader.

src/geometry/interaction_groups.rs
```diff
@@ -7,12 +7,19 @@
...
-/// In other words, interactions are allowed between two filter iff. the following condition is met:
+/// In other words, interactions are allowed between two filter iff. the following condition is met
+/// for [`Self::test_and`]:
 /// ```ignore
+/// ((self.0 >> 32) & rhs.0) != 0 && ((rhs.0 >> 32) & self.0) != 0
+/// ```
+/// or for [`Self::test_or`]:
+/// ```ignore
+/// ((self.0 >> 32) & rhs.0) != 0 || ((rhs.0 >> 32) & self.0) != 0
+/// ```
```
Vrixyz, line 22: The `iff.` followed by a condition with "or" is a bit difficult to follow, but that's probably me who is not used to `iff.`

Also, I'd probably use the user facing types (`InteractionTestMode::AND` rather than [`Self::test_and`]) and and/or link to [`Self::test`]:

src/geometry/interaction_groups.rs
```diff
@@ -56,24 +77,44 @@ impl InteractionGroups {
         self
     }
 
-    /// Check if interactions should be allowed based on the interaction memberships and filter.
+    /// Check if interactions should be allowed based on the interaction groups and mask.
```
Vrixyz, line 80: this naming change does not correspond to the properties of `InteractionGroup` anymore, if this is intentional, careful rename should appear throughout the crate (and probably a mention in the changelog, then userguides in rapier.rs), I'd be fine with considering that a follow-up.

src/geometry/interaction_groups.rs
Ughuuu: I would be inclined to not add it, as it's an ugly API (new_with_testmode), also it hides away this new functionality (people who use new still don't know they can use AND or OR).

CHANGELOG.md
Ughuuu: Moved to modified.

Ughuuu:
Fixed the doc test from CI.

Vrixyz (approved):

src/dynamics/coefficient_combine_rule.rs
```diff
@@ -27,8 +29,10 @@ impl CoefficientCombineRule {
 
         match effective_rule {
             0 => (coeff1 + coeff2) / 2.0,
-            1 => coeff1.min(coeff2),
+            1 => coeff1.min(coeff2).abs(),
```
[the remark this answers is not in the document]
Vrixyz: @sebcrozet why does this function takes u8 rather than `CoefficientCombineRule` ?

src/geometry/interaction_groups.rs
```diff
@@ -23,25 +28,39 @@ pub struct InteractionGroups {
...
+
+#[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, Default)]
+/// Specifies which method should be used to test interactions
+pub enum InteractionTestMode {
+    /// Use [`InteractionGroups::test_and`].
+    #[default]
+    AND,
+    /// Use [`InteractionGroups::test_or`].
+    OR,
```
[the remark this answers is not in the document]
Vrixyz: That would be caught with this clippy lint:

https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

with additional options in `clippy.toml` :

```
avoid-breaking-exported-api = false
upper-case-acronyms-aggressive = true
```
 
Unfortunately, This option also warns against `SAPRegion` and similar (+ `TOIEntry`, `CCDSolver`...) . We could suppress it for those specifics.

src/dynamics/coefficient_combine_rule.rs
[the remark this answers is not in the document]
Ughuuu: Not sure about the godot use case, this is just the exact code they have right now, so for parity I have that. Their use case is that they also have negative coefficients, and use this with absorbent field.
https://github.com/godotengine/godot/blob/0eadbdb5d0709e4e557e52377fa075d3e2f0ad1f/scene/resources/physics_material.h#L57
Then they do abs here, so the negative one becomes positive:
https://github.com/godotengine/godot/blob/0eadbdb5d0709e4e557e52377fa075d3e2f0ad1f/modules/godot_physics_3d/godot_body_pair_3d.cpp#L259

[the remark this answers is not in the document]
Ughuuu: Should i add a comment in code around that match explaining that the coefficient can be negative in some cases?

src/geometry/interaction_groups.rs
```diff
@@ -59,21 +84,44 @@ impl InteractionGroups {
...
+    pub const fn test_or(self, rhs: Self) -> bool {
+        (self.memberships.bits() & rhs.filter.bits()) != 0
+            || (rhs.memberships.bits() & self.filter.bits()) != 0
+    }
+
+    /// Check if interactions should be allowed based on the interaction memberships and filter.
+    ///
+    /// See [`InteractionTestMode`] for more info.
+    #[inline]
+    pub fn test(self, rhs: Self) -> bool {
```
Vrixyz, line 108: I had to remove the `const` due to `rhs` not being const. Is it bad?

Ughuuu: Looks good to me.

Vrixyz: I could fix it :tada:

sebcrozet (changes requested):
Thank you for this PR. Just a nitpick and it’s good to go.

sebcrozet:
Thank you!

Issue #734, opened 2024-09-20
Narrow phase incorrect incremental state update upon collider reparenting


hpmv:
When reparenting a collider to another rigidbody, the narrow phase code does not handle the reparenting correctly:
* Two colliders' interactions are ignored if they belong to the same parent rigidbody, which makes sense. However, this is only checked when **adding** the collider pair to the narrow phase (to either the interaction or contact graph), in `add_pair`.
* As a result, if two colliders used to be in the same parent but one is reparented, narrow phase ignores their collisions until some other mechanism notifies it of the pair again. The symptom of this is that if at the moment of reparenting the two colliders are already overlapping, narrow phase completely ignores the penetration until they are moved apart and have a chance to re-collide.
* On the other hand, if two colliders used to be in different parents but are reparented to the same rigidbody, narrow phase does not correctly ignore their collisions. The symptom of this is that if two objects are already overlapping (via e.g. squeezing each other) when the reparenting occurs, and if the positions of the two colliders in the rigidbody are also overlapping, then narrow phase continues to treat the overlapping as normal collisions and tries to resolve them, resulting in spontaneous accelerations of the parent rigidbody.

To reproduce (sorry don't have time to write a test case right now but here's exactly what to do):
* First issue: make two overlapping colliders in one rigidbody. Simulate. Then reparent one collider to a new rigidbody while keeping global position. Simulate. The two colliders will remain overlapping.
* Second issue: make two colliders in two separate rigidbodies that overlap. Simulate but without resolving the overlapping completely. Then reparent one collider to the other's rigidbody, keeping the global position. The new rigidbody will spontaneously accelerate.

To fix, we should handle collider parent change correctly in `handle_user_changes_on_colliders` in `narrow_phase.rs`, but I'm not sure if this can tackle the first issue, because when the pair was rejected earlier, narrow phase does not know of the pair. The second case can be correctly tackled this way by removing the pair that is no longer supposed to be in the graphs. But there may be other implications of parent change that I'm not aware of.

As a workaround, instead of reparenting I am removing and re-adding the same collider. This solves both issues.

Vrixyz:
Thanks for the detailed reproduction steps, I'm reproducing the first scenario in https://github.com/dimforge/rapier/pull/742

Vrixyz:
[...]

PR #742, opened 2024-09-26
Incorrect narrow_phase collisions after using ColliderSet::set_parent

Vrixyz:
- Follow up to #734

This PR adds 2 tests: 1 for each scenario.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Fix `KinematicCharacterController::move_shape` not respecting parameters `max_slope_climb_angle` and `min_slope_slide_angle`.
 - Improve ground detection reliability for `KinematicCharacterController`. (#715)
 - Fix wasm32 default values for physics hooks filter to be consistent with native: `COMPUTE_IMPULSES`.
-- Fix `set_parent` when ongoing collisions should be affected (#742):
+- Fix changing a collider parent when ongoing collisions should be affected (#742):
   - Fix collisions not being removed when a collider is parented to a rigidbody while in collision with it.
   - Fix collisions not being added when the parent was removed while intersecting a (previously) sibling collider.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - Fix `KinematicCharacterController::move_shape` not respecting parameters `max_slope_climb_angle` and `min_slope_slide_angle`.
 - Improve ground detection reliability for `KinematicCharacterController`. (#715)
 - Fix wasm32 default values for physics hooks filter to be consistent with native: `COMPUTE_IMPULSES`.
+- Fix `set_parent` when ongoing collisions should be affected (#742):
+  - Fix collisions not being removed when a collider is parented to a rigidbody while in collision with it.
+  - Fix collisions not being added when the parent was removed while intersecting a (previously) sibling collider.
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


src/geometry/narrow_phase.rs
```diff
@@ -455,12 +455,34 @@ impl NarrowPhase {
                             }
                         }
                     }
-
+                    // Remove the new parent from the collision graphs
```
Vrixyz, line 458:
```suggestion
                    // Remove the new parent from the contact and intersection graphs
```

Also, shouldn´t we remove that from graph_indices too?

src/geometry/narrow_phase.rs
```diff
@@ -1161,3 +1183,246 @@ impl NarrowPhase {
...
+
+        /* Create body 2, which will contain collider 2 at first. */
+        let rigid_body_2 = RigidBodyBuilder::dynamic()
+            .translation(vector![0.0, 0.0, 0.0])
+            .build();
+        let body_2_handle = rigid_body_set.insert(rigid_body_2);
+
+        /* Create collider 2. Parent it to rigid body 1. */
+        //let collider_2_handle =
+        //    collider_set.insert_with_parent(collider.build(), body_2_handle, &mut rigid_body_set);
```
Vrixyz, line 1336: to remove

src/geometry/narrow_phase.rs
```diff
@@ -1161,3 +1183,246 @@ impl NarrowPhase {
...
+        assert!(
+            rigid_body_set
+                .get(body_1_handle)
+                .unwrap()
+                .position()
+                .translation
+                .vector
+                .magnitude()
+                // TODO: this is probably a way too big value to test, consider lowering it.
+                // In the meantime, this proves that current behaviour is incorrect.
```
Vrixyz, line 1423: todo left here

src/geometry/narrow_phase.rs
```diff
@@ -1161,3 +1183,246 @@ impl NarrowPhase {
...
+
+        /* Parent collider 2 to body 2. */
+        collider_set.set_parent(collider_2_handle, Some(body_2_handle), &mut rigid_body_set);
+        narrow_phase.add_pair(
+            &collider_set,
+            &ColliderPair {
+                collider1: collider_2_handle,
+                collider2: collider_1_handle,
+            },
+        );
```
Vrixyz, line 1274: This is an internal api which we can't advocate for. a proper fix is still needed.

crates/rapier3d/Cargo.toml
```diff
@@ -69,7 +68,7 @@ required-features = ["dim3", "f32"]
 
 [dependencies]
 vec_map = { version = "0.8", optional = true }
-instant = { version = "0.1", features = ["now"], optional = true }
+web-time = { version = "1.1", optional = true }
```
Vrixyz, line 71: why is that showing here ? :thinking:

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 - Fix `KinematicCharacterController::move_shape` not respecting parameters `max_slope_climb_angle` and `min_slope_slide_angle`.
 - Improve ground detection reliability for `KinematicCharacterController`. (#715)
 - Fix wasm32 default values for physics hooks filter to be consistent with native: `COMPUTE_IMPULSES`.
+- Fix `set_parent` when ongoing collisions should be affected (#742):
```
Vrixyz, line 10:
```suggestion
- Fix changing a collider parent when ongoing collisions should be affected (#742):
```

sebcrozet (approved):
Looks good, thanks! I took the liberty of making the following changes:
- Removed debug print statements.
- Changed the test a bit to test two consecutive reparenting.
- Don’t rewove the pairs from the graph upon reparenting. I believe this isn’t needed because the collision information will be automatically cleared by the contact/intersection update function if needed.

PR #757, opened 2024-11-05
Fixed some typos.

PieroV:
There were some typos in `collider.rs` ("along along the y axis").
I removed the second along and capitalized Y to be more coherent with `src/dynamics/rigid_body_components.rs` (but I didn't add the various ` the latter has, I can do it if you want me to).


Vrixyz:
Thanks!

PR #762, opened 2024-11-25
Fixed index out of bounds at body_jacobians

GhostMinerPlus:
image
My program panicked at here, so I think it may be need to grow a bigger buffer in `Multibody::append`.


GhostMinerPlus:
Ignoring this, I found a new reason.

PR #767, opened 2024-12-05
Remove instant dependency in favor of web-time

Vrixyz:


src/counters/timer.rs
```diff
@@ -46,18 +50,18 @@ impl Timer {
...
-            self.start = Some(instant::now());
+            self.start = Some(Instant::now());
         }
     }
 
     /// The measured time between the last `.start()` and `.pause()` calls.
     pub fn time(&self) -> f64 {
-        self.time
+        self.time.as_secs_f64() / 1000.0
     }
```
Vrixyz, line 60: Is this ok ?

- https://github.com/rust-lang/rust/issues/122451

Vrixyz: :face_exhaling: :man_facepalming:  `* 1000`

src/counters/timer.rs
```diff
@@ -46,18 +49,18 @@ impl Timer {
...
         {
-            self.start = Some(instant::now());
+            self.start = Some(web_time::Instant::now());
         }
     }
 
     /// The measured time between the last `.start()` and `.pause()` calls.
     pub fn time(&self) -> f64 {
-        self.time
+        self.time.as_secs_f64() * 1000.0
```
Vrixyz, line 58: Is this ok ?

- https://github.com/rust-lang/rust/issues/122451

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 - `InteractionGroups` default value for `memberships` is now `GROUP_1` (#706)
 - `ImpulseJointSet::get_mut` has a new parameter `wake_up: bool`, to wake up connected bodies.
+- Removed unmaintained `instant` in favor of `web-time`.
```
sebcrozet, line 21: This also needs to indicate that the wasm-bindgen dep has been removed because it's no longer needed.

src/counters/timer.rs
sebcrozet: Let's return the `Duration` here instead.
And add a method `time_ms()` which returns this value.

src/counters/timer.rs
```diff
@@ -46,18 +49,18 @@ impl Timer {
...
     }
 }
 
 impl Display for Timer {
     fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
-        write!(f, "{}s", self.time)
+        write!(f, "{}s", self.time.as_secs_f32())
```
sebcrozet, line 64:
```suggestion
        write!(f, "{}ms", self.time_ms())
```

sebcrozet (changes requested):

sebcrozet (approved):

PR #770, opened 2024-12-06
Parry update (0.18?)

Vrixyz:
- Follow up to https://github.com/dimforge/parry/pull/273#issuecomment-2496130758
- blocked by https://github.com/dimforge/parry/pull/293
  - Waiting for a parry release

See following trace with example 2d 100 x 100 balls:

- :yellow_square: this trace is master (traverse_best_first)
- :red_square: external trace is this PR (traverse_depth_first)

image

### DIFF
--- a/crates/rapier_testbed2d-f64/Cargo.toml
+++ b/crates/rapier_testbed2d-f64/Cargo.toml
@@ -34,7 +34,7 @@ default = ["dim2"]
 dim2 = []
 parallel = ["rapier/parallel", "num_cpus"]
 other-backends = ["wrapped2d"]
-profiler_ui = [] #["dep:puffin_egui", "profiling/profile-with-puffin"]
+profiler_ui = ["dep:puffin_egui", "profiling/profile-with-puffin"]
 # See https://github.com/dimforge/rapier/issues/760.
 unstable-puffin-pr-235 = []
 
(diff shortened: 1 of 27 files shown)


crates/rapier_testbed2d-f64/Cargo.toml
```diff
@@ -34,12 +34,12 @@ default = ["dim2"]
...
 parallel = ["rapier/parallel", "num_cpus"]
 other-backends = ["wrapped2d"]
-profiling = ["dep:puffin_egui", "profiling/profile-with-puffin"]
+profiler_ui = ["dep:puffin_egui", "profiling/profile-with-puffin"]
 # See https://github.com/dimforge/rapier/issues/760.
 unstable-puffin-pr-235 = []
 
 [package.metadata.docs.rs]
-features = ["parallel", "profiling"]
+features = ["parallel", "profiler_ui"]
```
Vrixyz, line 42: this feature change name is to avoid naming a feature the same as a dependency, + better represent what this does.

Vrixyz: I'm not entirely sure, but I encountered problems with enabling `profiling/profile-with-tracy` without that change (because you cannot profile with 2 backends (puffin+tracy)), I imagine that enabled `profiling` too, so that feature name change would have fixed it.

sebcrozet:
Thanks!

Issue #370, opened 2022-07-23
Document whether debug colors are HSLA or something else


martin-t:
The [doc comment on DebugColor](https://github.com/dimforge/rapier/blob/master/src/pipeline/debug_render_pipeline/debug_render_style.rs#L5) says that it uses HSLA but [DebugRenderBackend::draw_line](https://github.com/dimforge/rapier/blob/master/src/pipeline/debug_render_pipeline/debug_render_backend.rs#L34) takes `[f32; 4]` so it's not clear from the docs what format the color is meant to be. If rapier uses HSLA everywhere, it should probably use `DebugColor` consistently everywhere.

There's a couple more places in code where i saw `[f32; 4]` (such as `sleep_color_multiplier`).

sebcrozet:
Yes, they are all HSLA by default. I agree that we should use `DebugColor` everywhere.

MaxBittker:
these docs currently use them as rgba:
https://rapier.rs/docs/user_guides/javascript/getting_started_js/

PR #774, opened 2024-12-14
Removed hardcoded values for DebugColor in various places

legendofa:
This PR aims to fix issue: https://github.com/dimforge/rapier/issues/370
I have tested these changes according to the [contribution guide](https://github.com/dimforge/rapier/blob/master/CONTRIBUTING.md) and did not find anything unusual.
Please let me know if there are other things that i need to consider as well :)


Vrixyz:
Thanks

PR #782, opened 2025-01-08
Release v0.23.0

sebcrozet:
### Fix

- The broad-phase region key has been replaced by an i64 in the f64 version of rapier, increasing the range before panics occur.
- Fix `BroadphaseMultiSap` not being able to serialize correctly with serde_json.
- Fix `KinematicCharacterController::move_shape` not respecting parameters `max_slope_climb_angle` and `min_slope_slide_angle`.
- Improve ground detection reliability for `KinematicCharacterController`. (#715)
- Fix wasm32 default values for physics hooks filter to be consistent with native: `COMPUTE_IMPULSES`.

### Added

- `RigidBodySet` and `ColliderSet` have a new constructor `with_capacity`.
- Use `profiling` crate to provide helpful profiling information in different tools.
  - The testbeds have been updated to use `puffin_egui`

### Modified

- `InteractionGroups` default value for `memberships` is now `GROUP_1` (#706)
- `ImpulseJointSet::get_mut` has a new parameter `wake_up: bool`, to wake up connected bodies.
- Removed unmaintained `instant` in favor of `web-time`. This effectively removes the `wasm-bindgen` transitive dependency as it's no longer needed.
- Significantly improve performances of `QueryPipeline::intersection_with_shape`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
-## Unreleased
+## v0.23.0 (08 Jan 2024)
 
 ### Fix
 
-- The region key has been replaced by an i64 in the f64 version of rapier, increasing the range before panics occur.
+- The broad-phase region key has been replaced by an i64 in the f64 version of rapier, increasing the range before panics occur.
 - Fix `BroadphaseMultiSap` not being able to serialize correctly with serde_json.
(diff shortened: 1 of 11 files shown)


sebcrozet:
Publish dry-run failure can be ignored because of the testbed depending on the unpublished rapier v0.23.

PR #789, opened 2025-01-14
Add joint_natural_frequency and joint_damping_ratio to GenericJoint.

### LINKED ISSUE #676: Add the possiblity to set joint_natural_frequency and joint_damping_ratio per joint
Right now one can set these 2 parameters per world, but not per joint. Having this setting available per joint would make the setting more versatile, making it possible to make other joints have soft limits, not just spring joints.

ThierryBerger:
- Fix #676

:warning: I'm not sure about multibodies, their `cfm_coeff` and `rhs_bias` computed from `joint_natural_frequency` and `joint_damping_ratio` is shared with all constraint solver methods, so we'd probably need a strategy to conciliate different values into one?

- I'd like to add example for that, but lack intuition on what values to input.

### DIFF
--- a/src/dynamics/joint/generic_joint.rs
+++ b/src/dynamics/joint/generic_joint.rs
@@ -248,6 +248,26 @@ pub struct GenericJoint {
     pub enabled: JointEnabled,
     /// User-defined data associated to this joint.
     pub user_data: u128,
+    /// > 0: the natural frequency used by the springs for joint constraint regularization.
+    ///
+    /// Increasing this value will make it so that penetrations get fixed more quickly.
+    /// (default: [`None`]).
+    ///
+    /// If [`None`],
+    /// [`IntegrationParameters::joint_natural_frequency`](crate::prelude::IntegrationParameters::joint_natural_frequency)
+    /// is used.
+    pub joint_natural_frequency: Option<Real>,
+
+    /// The fraction of critical damping applied to the joint for constraints regularization.
+    ///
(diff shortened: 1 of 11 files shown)


src/dynamics/solver/joint_constraint/joint_generic_constraint.rs
```diff
@@ -89,6 +89,13 @@ impl JointGenericTwoBodyConstraint {
             locked_axes,
         );
 
+        let joint_natural_frequency = joint
+            .joint_natural_frequency
+            .unwrap_or(params.joint_natural_frequency);
```
Vrixyz, line 94: TODO: remove `params.joint_natural_frequency` and `params.joint_damping_ratio`

src/dynamics/joint/generic_joint.rs
```diff
@@ -248,6 +248,18 @@ pub struct GenericJoint {
     pub enabled: JointEnabled,
     /// User-defined data associated to this joint.
     pub user_data: u128,
+    /// > 0: the natural frequency used by the springs for joint constraint regularization.
+    ///
+    /// Increasing this value will make it so that penetrations get fixed more quickly.
```
Vrixyz, line 253:
```suggestion
    /// Increasing this value will make it so that positioning errors get fixed more quickly.
```

sebcrozet:
Thank you for this PR. This has been added in master in the form of per-joint softness.

PR #794, opened 2025-02-02
Release rapier3d-meshloader v0.4.0

sebcrozet:
The currently published version (0.3.0) still relies on rapier 0.22 instead of 0.23.


sebcrozet:
CI failure us due to rapier3d-meshloader 0.4 not being published. Can be ignored.

PR #796, opened 2025-02-03
Physics context with arguments from PhysicsPipeline::step

ThierryBerger:
The objective is to offer a simplified route to most common usage of the library.

## Current status

- [x] create a `PhysicsContext`
- [x] Migrate all relevant tests to use it, to get a feeling on how it's useful
- [x] migrate other code
-  ~~add an example to **not** use that `PhysicsContext`~~
- [x] add more helper functions ? - maybe out of scope for this PR :thinking: 
- [x] changelog


src/physics_context.rs
```diff
@@ -0,0 +1,82 @@
...
+            &mut self.colliders,
+            &mut self.impulse_joint_set,
+            &mut self.multibody_joint_set,
+            &mut self.ccd_solver,
+            self.query_pipeline.as_mut(),
+            &self.hooks,
+            &self.events,
+        );
+    }
+}
```
Vrixyz, line 82: add shortcuts to creations (rigidbodies, colliders, both) ; (and consider queries in a future PR?)

src/physics_context.rs
```diff
@@ -0,0 +1,311 @@
+use parry::query::{NonlinearRigidMotion, ShapeCastOptions};
+
+use crate::prelude::*;
+
+/// Contains all arguments to be passed to [`PhysicsPipeline::step`]
+#[allow(missing_docs)]
+pub struct PhysicsContext {
```
Vrixyz, line 7: add serialize / deserialize (ignore physics pipeline, maybe more ; check docs)

src_testbed/testbed.rs
```diff
@@ -886,16 +899,16 @@ impl Testbed<'_, '_, '_, '_, '_, '_> {
 
             let controller = self.state.character_controller.unwrap_or_default();
             let phx = &mut self.harness.physics;
-            let character_body = &phx.bodies[character_handle];
-            let character_collider = &phx.colliders[character_body.colliders()[0]];
+            let character_body = &phx.context.bodies[character_handle];
+            let character_collider = &phx.context.colliders[character_body.colliders()[0]];
```
Vrixyz, line 903: to avoid adding those, maybe events and hooks should be directly in the harness ? or in a different property ?

sebcrozet:
Thank you for this PR. A similar `PhysicsWorld` struct has been added in recent versions of Rapier.

PR #803, opened 2025-03-05
Fix user changes handling

### LINKED ISSUE #802: Assertion failure: position change, enabled state change, and missing MODIFIED flag
// Error ------------------------------------------------

Ball altitude: 9.998297

thread 'main' panicked at /home/johannes/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rapier2d-0.23.0/src/geometry/broad_phase_multi_sap/sap_utils.rs:36:13:
assertion failed: e.floor() < RegionKey::MAX as Real
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

// What happened ----------------------------------------

1. Modify the translation or rotation (RigidBodyChanges::POSITION).
2. Disable the rigid body.
3. Step the simulation.

// Reproduce --------------------------------------------

In the for loop at the bottom.

```rust
use rapier2d::prelude::*;

fn main() {
    let mut rigid_body_set = RigidBodySet::new();
    let mut collider_set = ColliderSet::new();

    /* Create the ground. */
    let collider = ColliderBuilder::cuboid(100.0, 0.1).build();
    collider_set.insert(collider);

    /* Create the bouncing ball. */
    let rigid_body = RigidBodyBuilder::dynamic()
        .translation(vector![0.0, 10.0])
        .build();
    let collider = ColliderBuilder::ball(0.5).restitution(0.7).build();
    let ball_body_handle = rigid_body_set.insert(rigid_body);
    collider_set.insert_with_parent(collider, ball_body_handle, &mut rigid_body_set);

    /* Create other structures necessary for the simulation. */
    let gravity = vector![0.0, -9.81];
    let integration_parameters = IntegrationParameters::default();
    let mut physics_pipeline = PhysicsPipeline::new();
    let mut island_manager = IslandManager::new();
    let mut broad_phase = DefaultBroadPhase::new();
    let mut narrow_phase = NarrowPhase::new();
    let mut impulse_joint_set = ImpulseJointSet::new();
    let mut multibody_joint_set = MultibodyJointSet::new();
    let mut ccd_solver = CCDSolver::new();
    let mut query_pipeline = QueryPipeline::new();
    let physics_hooks = ();
    let event_handler = ();

    /* Run the game loop, stepping the simulation once per frame. */
    for _ in 0..200 {
        physics_pipeline.step(
            &gravity,
            &integration_parameters,
            &mut island_manager,
            &mut broad_phase,
            &mut narrow_phase,
            &mut rigid_body_set,
            &mut collider_set,
            &mut impulse_joint_set,
            &mut multibody_joint_set,
            &mut ccd_solver,
            Some(&mut query_pipeline),
            &physics_hooks,
            &event_handler,
        );

        let ball_body = &mut rigid_body_set[ball_body_handle];
        println!("Ball altitude: {}", ball_body.translation().y);

	// This causes the error:
	// Modify the translation (or rotation) (RigidBodyChanges::POSITION)
	// Modify enabled
        ball_body.set_translation(vector![0.0, 0.0], true);
        // ball_body.set_rotation(nalgebra::UnitComplex::new(1.0), true);
        ball_body.set_enabled(false);
    }
}
```

// Solution (maybe) -------------------------------------------

rapier/src/dynamics/rigid_body_components.rs
line 970

In the function update_positions in the RigidBodyColliders impl:

```rust
pub fn update_positions(
    &self,
    colliders: &mut ColliderSet,
    modified_colliders: &mut Vec<ColliderHandle>,
    parent_pos: &Isometry<Real>,
) {
for handle in &self.0 {
    // NOTE: the ColliderParent component must exist if we enter this method.
    let co = colliders.index_mut_internal(*handle);
    let new_pos = parent_pos * co.parent.as_ref().unwrap().pos_wrt_parent;

    if !co.changes.contains(ColliderChanges::MODIFIED) {
        modified_colliders.push(*handle);
    }

    // Set the modification flag so we can benefit from the modification-tracking
    // when updating the narrow-phase/broad-phase afterwards.
    co.changes |= ColliderChanges::POSITION;
    co.pos = ColliderPosition(new_pos);
}
}
```

The handle is only added to modified_colliders if it does not already contain ColliderChanges::MODIFIED.
However, I believe the ColliderChanges::MODIFIED flag must be inserted into the changes when pushing
the handle to modified_colliders.

rapier/src/pipeline/user_changes.rs

- line 141 and 142
    update_positions is called and colliders may be added to the modified_colliders

- line 154 and 170
    Collider handles are added to the modified_colliders if the ColliderChanges::MODIFIED is not set
    
If a collider is pushed to modified_colliders in the update_positions function without setting the MODIFIED flag,
it can be pushed to modified_colliders twice.

Solution:
rapier/src/dynamics/rigid_body_components.rs
in the update_positions function:

```rust
    if !co.changes.contains(ColliderChanges::MODIFIED) {
        co.changes |= ColliderChanges::MODIFIED; // Insert ColliderChangesMODIFIED

        modified_colliders.push(*handle);
    }
```

// Also -----------------------------------------------

rapier/src/geometry/collider_set.rs
line 69

In the funktion iter_mut

```rust
/// Iterates mutably through all the colliders on this set.
#[cfg(not(feature = "dev-remove-slow-accessors"))]
pub fn iter_mut(&mut self) -> impl Iterator<Item = (ColliderHandle, &mut Collider)> {
    self.modified_colliders.clear();
    let modified_colliders = &mut self.modified_colliders;
    self.colliders.iter_mut().map(move |(h, b)| {
        modified_colliders.push(ColliderHandle(h));
        (ColliderHandle(h), b)
    })
}
```

1. self.modified_colliders.clear() is called at the beginning, resetting the list.
2. Every collider handle is then pushed into modified_colliders inside the map function without first setting the MODIFIED flag.

```rust
/// Iterates mutably through all the colliders on this set.
#[cfg(not(feature = "dev-remove-slow-accessors"))]
pub fn iter_mut(&mut self) -> impl Iterator<Item = (ColliderHandle, &mut Collider)> {
    self.modified_colliders.clear();
    let modified_colliders = &mut self.modified_colliders;
    self.colliders.iter_mut().map(move |(h, b)| {
        b.changes |= ColliderChanges::MODIFIED; // Ensure the MODIFIED flag is set
        modified_colliders.push(ColliderHandle(h));
        (ColliderHandle(h), b)
    })
}
```

Vrixyz:
Fix #802

Note: Fixing those + adding tests is a good first step but we should thrive for making those mistakes more difficult to make altogether.

### DIFF
--- a/src/dynamics/rigid_body_components.rs
+++ b/src/dynamics/rigid_body_components.rs
@@ -1,7 +1,7 @@
 use crate::dynamics::MassProperties;
 use crate::geometry::{
     ColliderChanges, ColliderHandle, ColliderMassProps, ColliderParent, ColliderPosition,
-    ColliderSet, ColliderShape,
+    ColliderSet, ColliderShape, ModifiedColliders,
 };
 use crate::math::{
     AngVector, AngularInertia, Isometry, Point, Real, Rotation, Translation, Vector,
@@ -955,10 +955,10 @@ impl RigidBodyColliders {
     }
 
     /// Update the positions of all the colliders attached to this rigid-body.
-    pub fn update_positions(
+    pub(crate) fn update_positions(
         &self,
         colliders: &mut ColliderSet,
-        modified_colliders: &mut Vec<ColliderHandle>,
+        modified_colliders: &mut ModifiedColliders,
         parent_pos: &Isometry<Real>,
     ) {
         for handle in &self.0 {
@@ -966,13 +966,11 @@ impl RigidBodyColliders {
             let co = colliders.index_mut_internal(*handle);
             let new_pos = parent_pos * co.parent.as_ref().unwrap().pos_wrt_parent;
 
-            if !co.changes.contains(ColliderChanges::MODIFIED) {
-                modified_colliders.push(*handle);
-            }
-
             // Set the modification flag so we can benefit from the modification-tracking
             // when updating the narrow-phase/broad-phase afterwards.
-            co.changes |= ColliderChanges::MODIFIED | ColliderChanges::POSITION;
+            modified_colliders.push_once(*handle, co);
+
+            co.changes |= ColliderChanges::POSITION;
             co.pos = ColliderPosition(new_pos);
         }
     }
--- a/src/dynamics/rigid_body_components.rs
+++ b/src/dynamics/rigid_body_components.rs
@@ -972,7 +972,7 @@ impl RigidBodyColliders {
 
             // Set the modification flag so we can benefit from the modification-tracking
             // when updating the narrow-phase/broad-phase afterwards.
-            co.changes |= ColliderChanges::POSITION;
+            co.changes |= ColliderChanges::MODIFIED | ColliderChanges::POSITION;
             co.pos = ColliderPosition(new_pos);
         }
     }
--- a/src/geometry/collider_set.rs
+++ b/src/geometry/collider_set.rs
@@ -66,6 +66,7 @@ impl ColliderSet {
         self.modified_colliders.clear();
         let modified_colliders = &mut self.modified_colliders;
         self.colliders.iter_mut().map(move |(h, b)| {
+            b.changes |= ColliderChanges::MODIFIED;
             modified_colliders.push(ColliderHandle(h));
             (ColliderHandle(h), b)
         })
--- a/src/geometry/collider_set.rs
+++ b/src/geometry/collider_set.rs
@@ -1,15 +1,30 @@
 use crate::data::arena::Arena;
+use crate::data::{HasModifiedFlag, ModifiedObjects};
 use crate::dynamics::{IslandManager, RigidBodyHandle, RigidBodySet};
 use crate::geometry::{Collider, ColliderChanges, ColliderHandle, ColliderParent};
 use crate::math::Isometry;
 use std::ops::{Index, IndexMut};
+
+pub(crate) type ModifiedColliders = ModifiedObjects<ColliderHandle, Collider>;
+
+impl HasModifiedFlag for Collider {
+    #[inline]
+    fn has_modified_flag(&self) -> bool {
+        self.changes.contains(ColliderChanges::MODIFIED)
+    }
+
+    #[inline]
+    fn set_modified_flag(&mut self) {
+        self.changes |= ColliderChanges::MODIFIED;
+    }
+}
 
 #[cfg_attr(feature = "serde-serialize", derive(Serialize, Deserialize))]
 #[derive(Clone, Default, Debug)]
 /// A set of colliders that can be handled by a physics `World`.
 pub struct ColliderSet {
     pub(crate) colliders: Arena<Collider>,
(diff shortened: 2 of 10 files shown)


src/dynamics/rigid_body_components.rs
```diff
@@ -972,7 +972,7 @@ impl RigidBodyColliders {
 
             // Set the modification flag so we can benefit from the modification-tracking
             // when updating the narrow-phase/broad-phase afterwards.
-            co.changes |= ColliderChanges::POSITION;
+            co.changes |= ColliderChanges::MODIFIED | ColliderChanges::POSITION;
```
Vrixyz, line 975: This makes me wonder if we should change `ColliderChanges::POSITION` to  `const POSITION = 1 << 3 | 1 << 0;` (adding `ColliderChanges::MODIFIED` to it), this would clarify their relation and make mistakes harder to make.

src/geometry/collider_set.rs
```diff
@@ -66,6 +66,7 @@ impl ColliderSet {
         self.modified_colliders.clear();
         let modified_colliders = &mut self.modified_colliders;
         self.colliders.iter_mut().map(move |(h, b)| {
+            b.changes |= ColliderChanges::MODIFIED;
```
Vrixyz, line 69: I'm not sure this is correct ; the fix may be more around `https://github.com/Vrixyz/rapier/blob/607f0988596d346b6afa70e5cfbc8f0f3519e02c/src/geometry/collider.rs#L271` ; to set `ColliderChanges::MODIFIED` flag too (see https://github.com/dimforge/rapier/pull/803/files#r1980964215 comment for a longer term fix maybe ?)

And then add then refill `self.modified_colliders` only when applicable (check `changes` flag)

src/dynamics/rigid_body_components.rs
Vrixyz: This change is probably not possible because `MODIFIED` is used to check if a collider is part of the `modified_colliders` list. This seems difficult to maintain, I imagine searching for `modified_colliders.push`  should be able to lead to a `MODIFIED` bit to be added to that collider, but it's sometimes tricky to assess. 

I imagine changing `modified_colliders` to a `Hashmap` may be less performant than storing that in the bitfield, but I'm throwing the idea.

src/pipeline/physics_pipeline.rs
```diff
@@ -1011,4 +1011,66 @@ mod test {
...
+            println!("Ball altitude: {}", ball_body.translation().y);
+
+            // This causes the error:
+            // Modify the translation (or rotation) (RigidBodyChanges::POSITION)
+            // Modify enabled
+            ball_body.set_translation(vector![0.0, 0.0], true);
+            // ball_body.set_rotation(nalgebra::UnitComplex::new(1.0), true);
+            ball_body.set_enabled(false);
+        }
+    }
```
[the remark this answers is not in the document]
Johannes0021: I rewrote it again, set default values at the beginning, and ensured that the values change.

PR #817, opened 2025-03-31
Prevent underflow with saturating_sub(1) (#812)

### LINKED ISSUE #812: Possible underflow in num_islands and num_groups - Would saturating_sub(1) be better?
Info to saturating_sub:
Performs subtraction that saturates at the numeric bounds instead of overflowing.

////////////////////////////////////////////////////////////////////////////////////////////////////
// 'rapier/src/dynamics/island_manager.rs'
////////////////////////////////////////////////////////////////////////////////////////////////////

Test case (If we assume that num_islands is initialized with an empty vector.):
```rust
#[cfg(test)]
mod tests {

    use super::*;



    #[test]
    fn test_num_islands() {
        let manager = IslandManager::new();

        assert_eq!(manager.num_islands(), 0);
    }

}
```

Old function (line: 39):
```rust
pub(crate) fn num_islands(&self) -> usize {
    self.active_islands.len() - 1
}
```

New function with saturating_sub(1):
```rust
pub(crate) fn num_islands(&self) -> usize {
    self.active_islands.len().saturating_sub(1)
}
```

////////////////////////////////////////////////////////////////////////////////////////////////////

Line 248 could also be solved with saturating_sub(1):
```rust
let mut island_marker = self.stack.len().max(1) - 1;
```

To:
```rust
let mut island_marker = self.stack.len().saturating_sub(1);
```

////////////////////////////////////////////////////////////////////////////////////////////////////
// 'rapier/src/dynamics/solver/interaction_groups.rs'
////////////////////////////////////////////////////////////////////////////////////////////////////

Test case (If we assume that num_groups is initialized with an empty vector.)
Run with 'cargo test --features parallel':
```rust
#[cfg(feature = "parallel")]
#[cfg(test)]
mod tests {

    use super::*;



    #[test]
    fn test_num_groups() {
        let groups = ParallelInteractionGroups::new();

        assert_eq!(groups.num_groups(), 0);
    }

}
```

Old function (line: 61):
```rust
pub fn num_groups(&self) -> usize {
    self.groups.len() - 1
}
```

New function with saturating_sub(1):
```rust
pub fn num_groups(&self) -> usize {
    self.groups.len().saturating_sub(1)
}
```

Johannes0021:
Fixes #812

### DIFF
--- a/src/dynamics/island_manager.rs
+++ b/src/dynamics/island_manager.rs
@@ -37,7 +37,7 @@ impl IslandManager {
     }
 
     pub(crate) fn num_islands(&self) -> usize {
-        self.active_islands.len() - 1
+        self.active_islands.len().saturating_sub(1)
     }
 
     /// Update this data-structure after one or multiple rigid-bodies have been removed for `bodies`.
@@ -244,8 +244,8 @@ impl IslandManager {
         self.active_islands.clear();
         self.active_islands.push(0);
 
-        // The max avoid underflow when the stack is empty.
-        let mut island_marker = self.stack.len().max(1) - 1;
+        // saturating_sub(1) prevents underflow when the stack is empty.
+        let mut island_marker = self.stack.len().saturating_sub(1);
 
         // NOTE: islands containing a body with non-standard number of iterations won’t
         //       be merged with another island, unless another island with standard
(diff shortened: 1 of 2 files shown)


Vrixyz:
Thanks!

PR #823, opened 2025-04-17
feat: add support for Voxels collider

sebcrozet:
Depends on https://github.com/dimforge/parry/pull/336
This adds supports for Voxels colliders: a shape specific to colliders defined as a set of voxels. This automatically handles internal edges and supports dynamic addition/removal of voxels. It is a faster alternative, and with a lower memory footprint, than triangle meshes when the scene is only made of voxels.

This feature is still experimental. See what’s missing in the Parry PR description https://github.com/dimforge/parry/pull/336.

image
image


sebcrozet:
CI will fail until the parry PR is merged & released.

PR #825, opened 2025-04-18
Add a customizable global error handler

ThierryBerger:
This allows user to react to an engine error, be it panic or try to fix the problem.

Took inspiration from https://github.com/bevyengine/bevy/blob/main/crates/bevy_ecs/src/error/handler.rs.

## Current Status

Proof of concept is here, I'd like to add an example or test where user code is fixing an issue (removes a rigidbody ?)

More errors should probably be added in there.

Should this be behind a feature flag ?

no-std should be accounted for: be it allow it (custom oncelock), or not supporting it.

### DIFF
--- a/src/error_handler.rs
+++ b/src/error_handler.rs
@@ -1,0 +1,129 @@
+//! Error handling for rapier.
+//!
+//! A number of checks are in place to catch inconsistencies as soon as they occur,
+//! while it's best to make sure none of these errors occur, some may be recoverable
+//! through removing the root cause (incorrect shape or position for example).
+//!
+//! Setting [`GLOBAL_ERROR_HANDLER`] can help you as the end user to react to those errors.
+//!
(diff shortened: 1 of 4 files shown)


src/error_handler.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        let (tx, rx): (Sender<i32>, Receiver<i32>) = mpsc::channel();
+        // Set error handling
+        if GLOBAL_ERROR_HANDLER
+            .set(Box::new(move |error| {
+                println!("error: {:?}", error);
+                // TODO: add more context:
+                // - offending rigidbody ?
+                // - backtrace ?
+                assert!(tx.send(1).is_ok());
```
Vrixyz, line 99: TODO: send the error

sebcrozet:
Thank you for this PR. Closing in favor of future rework for error handling with more idiomatic propagation.

PR #834, opened 2025-05-14
CollisionPipeline updates collider when parent body changes

Kuderbe:
When inserting a Collider with a RigidBody as a parent, the collider doesn't get updated after the first step of the collision Pipeline. The Documentation states:

> If a collider is attached to a rigid-body, its position is automatically updated by the physics pipeline when a rigid-body is moved by the physics pipeline. If a change to the rigid-body position is made by the user then the collider position will be updated during the next timestep.

While this doesn't mention the CollisionPipeline I would expect the same behavior when it comes to position changes made by the user.

Currently those changes only get made once when the body is marked as modified but never a second time, as the flag never gets reset.

This Pull-Request adds a Unit test which tests for the expected behavior and a fix strongly inspired by the PhysicsPipeline.

I'm not a 100% sure that this doesn't open any new issues, however the unit tests all pass and the demos seem to work as expected. I'm obviously also open for different approaches for reaching the expected behavior. However I would like to avoid using the full PhysicsPipeline as I only need collision detection and would like to avoid the overhead of additional physics computations.


sebcrozet:
Thank you for this PR! Part of this fix has been integrated in the recent fixes on master.

PR #835, opened 2025-05-16
Release d0.26.0

sebcrozet:
### Modified

- Update to parry 0.21.0. This changes the initialization of `Voxels` colliders by removing the primitive geometry argument.
- This also fixes intersection checks with voxels, and force calculation between voxels and other voxels or compound shapes.


sebcrozet:
Publish dry-run failure is expected.

PR #846, opened 2025-06-12
Update testbeds to bevy to 0.16 and egui to 0.31

ThierryBerger:


crates/rapier_testbed2d-f64/Cargo.toml
```diff
@@ -67,31 +67,35 @@ indexmap = { version = "2", features = ["serde"] }
 
 # Dependencies for native only.
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-bevy = { version = "0.15", default-features = false, features = [
+bevy = { version = "0.16", default-features = false, features = [
     "bevy_asset",
     "bevy_winit",
     "bevy_window",
     "x11",
+    "wayland",
```
Vrixyz, line 75: This fixes experience for linux users but crashes ci, not worth having, or maybe only for dev-dependencies ? or upstream fix?

sebcrozet:
Thank you for this PR. The testbed now uses kiss3d instead of bevy.

PR #851, opened 2025-07-03
Fix web_time usage for Timer

pmhpereira:
rapier3d failed to compile for wasm32 with the `profiler` feature flag enabled.
This PR fixes it, making the `profiler` flag work as expected across any target_arch.

### DIFF
--- a/src/counters/timer.rs
+++ b/src/counters/timer.rs
@@ -3,12 +3,16 @@ use std::{
     time::Duration,
 };
 
+#[cfg(feature = "profiler")]
+use web_time::Instant;
+
 /// A timer.
 #[derive(Copy, Clone, Debug, Default)]
 pub struct Timer {
     time: Duration,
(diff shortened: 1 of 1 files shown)


src/counters/timer.rs
```diff
@@ -3,19 +3,24 @@ use std::{
...
 };
 
+#[cfg(feature = "profiler")]
+use web_time::Instant;
+
 /// A timer.
 #[derive(Copy, Clone, Debug, Default)]
 pub struct Timer {
     time: Duration,
     #[allow(dead_code)] // The field isn’t used if the `profiler` feature isn’t enabled.
```
Vrixyz, line 13: So this may be removed then ?

Vrixyz (approved):

src/counters/timer.rs
pmhpereira: Done.

sebcrozet:
Thanks!

PR #861, opened 2025-07-18
Replace crossbeam channel with std::sync::mpsc

### LINKED ISSUE #828: Consider using std::sync::mpsc instead of crossbeam-channel
How about using `std::sync::mpsc` instead of `crossbeam-channel`?

As far as I know, the channel code from `crossbeam-channel` has been merged into the standard library's `std::sync::mpsc`.

Since I would prefer not to install additional crates just to use Rapier, I'd like to suggest using the standard library implementation, unless there is a specific need or feature from `crossbeam-channel` that Rapier explicitly relies on.

namse:
Replaced all uses of `crossbeam::channel` with `std::sync::mpsc` to reduce the dependency footprint.

## Changes
- Modified channel imports from `crossbeam::channel::{Sender, Receiver}` to `std::sync::mpsc::{Sender, Receiver}`
- Changed `crossbeam::channel::unbounded()` to `std::sync::mpsc::channel()`
- Removed `crossbeam = "0.8"` dependency from 8 Cargo.toml files
- Removed `pub extern crate crossbeam;` from src/lib.rs
- Updated doc comments to remove crossbeam mentions

As discussed in #828, crossbeam is only used for its channel implementation. Since std::sync::mpsc provides equivalent functionality for this use case, we can use the standard library instead.

Fixes #828

### DIFF
--- a/crates/rapier2d-f64/Cargo.toml
+++ b/crates/rapier2d-f64/Cargo.toml
@@ -73,7 +73,6 @@ parry2d-f64 = "0.22.0-beta.1"
 simba = "0.9"
 approx = "0.5"
 rayon = { version = "1", optional = true }
-crossbeam = "0.8"
 arrayvec = "0.7"
 bit-vec = "0.8"
 rustc-hash = "2"
--- a/crates/rapier2d/Cargo.toml
+++ b/crates/rapier2d/Cargo.toml
@@ -74,7 +74,6 @@ parry2d = "0.22.0-beta.1"
 simba = "0.9"
 approx = "0.5"
 rayon = { version = "1", optional = true }
-crossbeam = "0.8"
 arrayvec = "0.7"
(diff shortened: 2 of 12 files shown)


Ptrskay3:
Hey!

I just want to mention that feature-wise `std::sync::mpsc` may be equivalent, but there is a difference. `crossbeam::channel::Receiver` [implements `Sync`](https://docs.rs/crossbeam/latest/crossbeam/channel/struct.Receiver.html#impl-Sync-for-Receiver%3CT%3E), but `std::sync::mpsc::Receiver` is [`!Sync`](https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.Receiver.html#impl-Sync-for-Receiver%3CT%3E), and this is a breaking change. This change is blocking our upgrade path (or requires a significant re-thinking of our current structure).

PR #864, opened 2025-07-24
add a flag `first_tick` to `ContactForceEvent `

bcolloran:
This adds a flag "first_tick" to `ContactForceEvent`, that is true if this is the first tick on which a contact force has been applied by the corresponding `ContactPair`.

This seems to be a tidy solution to the issue I described in this discord issue:
https://discord.com/channels/507548572338880513/747935665076830259/1397751543700262934

Please let me know if you have any feedback or suggestions, or if you see anything incorrect. Thanks!


bcolloran:
Hi @sebcrozet, any chance to get a review on this little teeny change :-) 

This is useful for example for playing a sound just once at the moment when a collision starts.

Happy to do the implementation of a different approach if you have another suggested way of doing this! 

Thanks much!

sebcrozet:
Thank you for this PR. The proposed approach is no longer applicable to the current version of Rapier, and has the issue that a new contact doesn’t imply it’s the first time we exceed the threshold. This will be implemented in #981.

PR #871, opened 2025-08-07
remove accidental new dependencies

Timbals:
As far as I can tell, these new dependencies aren't used and are probably just left over from experimenting.


sebcrozet:
Good catch, thank you!

PR #873, opened 2025-08-30
Fix unsound uses of `get_unchecked_mut`

vdrn:
There were multiple unsound uses of `get_unchecked_mut`.

**That function does not allow you to get multiple mutable references into a slice.**
**It only skips bounds-checking, nothing else.**

You can run this code with `cargo miri run` to see the issue:
```rust
fn main() {
    let mut v: Vec<i32> = vec![1, 2];
    unsafe {
        let a = v.get_unchecked_mut(0) as *mut i32;
        let b = v.get_unchecked_mut(1) as *mut i32;
        println!("{}, {}", *a, *b);
    }
}
```

Since `1.86`, stable Rust provides the exact functions you need:
`get_disjoint_mut` and `get_disjoint_unchecked_mut`.

Notes:
- I've tried to make this pull request as non-intrusive as possible, but IMO `IndexMut2` Trait is redundant and should probably be removed.
- In `get_mut_with_parent`, I've kept the same error message for "Internal error: circular rigid body dependency. ", but that probably should just be `get_disjoint_mut(..).unwrap()`

### DIFF
--- a/src/data/graph.rs
+++ b/src/data/graph.rs
@@ -147,8 +147,7 @@ fn index_twice<T>(arr: &mut [T], a: usize, b: usize) -> Pair<&mut T> {
     } else {
         // safe because a, b are in bounds and distinct
         unsafe {
-            let ar = &mut *(arr.get_unchecked_mut(a) as *mut _);
-            let br = &mut *(arr.get_unchecked_mut(b) as *mut _);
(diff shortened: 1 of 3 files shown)


sebcrozet:
Thank you for this PR. Most (hopefully all) unsoundness has been adressed in the latest rapier version. And we are now also running miri checks in CI.

PR #876, opened 2025-09-04
feat: solver improvements + release v0.29.0

sebcrozet:
This PR reworks the constraints solver in order to both simplify and optimize it.

### Simplifications
The simplifications aim to make it easier the future to optimize the solver and experiment with new approaches. This allows the removal of large chunks of code that were generally mathematically redundant.

- Removed all one-body variants of all constraints (joints and contacts). These where constraint that involved only one dynamic rigid-body. They are now simulated with two-body constraints.
- When SIMD is enabled, all contact constraints now use the SIMD two-body constraint (instead of having some contact constraints be SIMD-accelerated, and some defaulting to the scalar implementation).
- Kinematic rigid-body velocities are now integrated the same way as dynamic rigid-bodies by the velocity solver.
- Refactored gather/scatter patterns for solver bodies and constraints so that all constraints reuses the same code.
- Angular inertia tensors are no longer represented by their square root. This was a generally efficient optimization that is now significantly less impactful due to the other optimizations. Removing that optimization simplifies the solver significantly by unifying the way linear and angular velocities are handled. This also allows simplifications on the way kinematic bodies are simulation.
- Removed support for the legacy PGS solver. This allow us to simplify the main substep loop and remove a couple of code paths.

### Optimizations
- Reworked the solver body gather/scatter patterns to leverage SIMD transpositions to switch between SoA and AoS formats (note that https://github.com/Lokathor/wide/pull/211 will be needed to get all the performance benefits from that rework on arm and wasm).
- Linearized angular velocity integration to avoid trigonometric functions at each simulation substep.
- (3D only:) Implemented a simplified friction model that only generates a single Coulomb friction + one angular constraint per contact manifold (similar to PhysX’s "patch friction" system). This is a simplification over the traditional Coulomb friction but (1) yields the same result for a single box sliding on a flat surface and (2) is generally indistinguishable from regular coulomb friction for gaming purposes. This new model is now enabled by default but can be changed back globally with `IntegrationParameters::friction_model`.
- (2D only:) I realized that usage of explicit SIMD intrinsics were actually disabled on `rapier2d`, even when its `simd-stable` feature was enabled. This resulted in a performance loss due to relying on the compiler’s auto-vectorization rather than the manual intrinsics calls… This is now fixed.

### Features
- (3D only:) Implemented gyroscopic forces. They can be enabled with `RigidBodyBuilder::gyroscopic_forces_enabled` or `RigidBody::enable_gyroscopic_forces`.

### Fixes
- Fix friction having an abnormally weak effect on dynamic rigid-bodies resting on a moving kinematic rigid-body.

### Results

Overall performance improvements are very dependent on the content of the simulation. The general speedup on contacts-heavy scenes is of about 25%. The timings below are given assuming that the `simd-stable` feature is enabled (and compilation is in release mode of course).

###### Keva tower (3D)
Total step time went down from 20ms to 14ms.
**Before:**
image

**Now:**
image

###### Tower of blocks (2D)

Total step time went down from 25ms to 18ms.
**Before:**
image
**Now:**
image

### DIFF
--- a/src/dynamics/solver/contact_constraint/any_contact_constraint.rs
+++ b/src/dynamics/solver/contact_constraint/any_contact_constraint.rs
@@ -9,19 +9,19 @@ use crate::prelude::ContactManifold;
 
 #[derive(Debug)]
 pub enum AnyContactConstraintMut<'a> {
-    GenericTwoBodies(&'a mut GenericContactConstraint),
-    SimdTwoBodiesCoulomb(&'a mut ContactWithCoulombFriction),
+    Generic(&'a mut GenericContactConstraint),
+    WithCoulombFriction(&'a mut ContactWithCoulombFriction),
     #[cfg(feature = "dim3")]
-    SimdTwoBodiesTwist(&'a mut ContactWithTwistFriction),
+    WithTwistFriction(&'a mut ContactWithTwistFriction),
 }
 
 impl AnyContactConstraintMut<'_> {
     pub fn remove_bias(&mut self) {
         match self {
-            Self::GenericTwoBodies(c) => c.remove_cfm_and_bias_from_rhs(),
-            Self::SimdTwoBodiesCoulomb(c) => c.remove_cfm_and_bias_from_rhs(),
+            Self::Generic(c) => c.remove_cfm_and_bias_from_rhs(),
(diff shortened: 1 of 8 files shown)


awe719:
Nice 🙂

PR #887, opened 2025-10-26
Expose ColliderSet::{take_modified, take_removed}

Ralith:
These are practically necessary when working without a PhysicsPipeline.

A larger change could avoid exposing `ModifiedObjects`, but it seems like it might be useful as well.

Alternatively, `BroadPhaseBvh::update` could be modified to pull this data from the `ColliderSet` itself. In that case the public interface could be narrowed to a method on `ColliderSet` to clear the modified/removed sets, but this would make it impossible for a user to inject their own modified/removed handles. I'm not sure if that's useful.

### DIFF
--- a/src/data/mod.rs
+++ b/src/data/mod.rs
@@ -2,7 +2,7 @@
 
 pub use self::arena::{Arena, Index};
 pub use self::coarena::Coarena;
-pub(crate) use self::modified_objects::{HasModifiedFlag, ModifiedObjects};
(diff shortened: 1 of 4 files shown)


sebcrozet:
Thank you! I don’t _love_ that change, as it increases the risks of user errors, but I don’t have a better alternative yet. Let’s use that until the whole "use BroadPhase independently from PhysicsPipeline" story gets better.

PR #894, opened 2025-11-02
Fix issue where set_enabled had no effect

Deniskore:
This PR fixes an issue where calling `set_enabled(false)` had no effect. There may be a more appropriate solution to this issue, and feedback from maintainers is needed.

Related issues:
1. [dimforge/bevy_rapier#680](https://github.com/dimforge/bevy_rapier/pull/680)  
2. https://github.com/dimforge/bevy_rapier/pull/678#issuecomment-3397808035
3. [dimforge/rapier.js#344](https://github.com/dimforge/rapier.js/issues/344)

Reproduction:
Steps to reproduce the issue are clearly described in the [Update to Bevy 0.17.2 PR](https://github.com/dimforge/bevy_rapier/pull/680).

### DIFF
--- a/src/geometry/narrow_phase.rs
+++ b/src/geometry/narrow_phase.rs
@@ -831,6 +831,10 @@ impl NarrowPhase {
                     // No update needed for these colliders.
                     return;
                 }
+                if !co1.is_enabled() || !co2.is_enabled() {
+                    pair.clear();
+                    break 'emit_events;
+                }
                 if co1.parent.map(|p| p.handle) == co2.parent.map(|p| p.handle) && co1.parent.is_some()
                 {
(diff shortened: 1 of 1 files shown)


Deniskore:
@sebcrozet I rebased my branch to the latest changes in the main branch, could you rerun CI?

sebcrozet:
I found the root of the problem. Thank you for the PR @Deniskore, but it will be replaced by #900.

Deniskore:
> I found the root of the problem. Thank you for the PR @Deniskore, but it will be replaced by #900.

The fix works. Could you tell when you will be able to release the new version so that we can update the plugin for Bevy as well?

PR #895, opened 2025-11-07
feat: persistent islands + manifold reduction

sebcrozet:
This PR adds contact manifold reduction, ensuring that each contact manifold only result in up to 4 solver contacts (in 3D). This simplifier the solver code further by ensuring a 1-to-1 matching between solver contact manifolds and contact constraints. Two manifold reductions strategies are implemented:
- One "naive" approach that takes the deepest contact, the one furthest to it, and the two contacts furthest to the segment formed by the first two.
- A more complex approach (albeit similar to the first one) that is a port of Bepu’s manifold reduction. This has been implemented for comparison purpose.

In practice, I selected the first approach as it appeared to work well with less operations than the more complex one. I’m keeping the second approach around for now for additional future experiments.

----

This PR also adds persistent islands. Instead of recalculating the simulation islands at each simulation step, both awake and sleeping islands are persisted across frames and updated incrementally. Whenever a rigid-body desires to sleep, it triggers (in some future frame) a connected-component traversal of the contact/joint graphs in order to attempt to extract a sleeping island.

In addition, an islands optimizer (that could use a bit more tuning) has been implemented to ensure (as long as it doesn’t affect correctness) that islands are neither too small nor too big. This results in generally better performances due to better cache utilisation. The island optimizer runs a small amount of work at each frame, progressively balancing the islands sizes.

The persistent islands optimization results in a performance improvement between 10% to 20% depending on the stress test. The graph below are examples of perf improvements (blue is current `master`, red is this PR) on some of our stress tests (`all_benchmarks3`):
balls-times
capsules-times
stacks-times

### DIFF
--- a/src/dynamics/island_manager/island.rs
+++ b/src/dynamics/island_manager/island.rs
@@ -77,7 +77,6 @@ impl IslandManager {
             );
             rb.ids.active_island_id = new_island_id;
             rb.ids.active_set_id = id;
-            rb.ids.active_set_offset = id as u32;
 
             new_island.additional_solver_iterations = new_island
                 .additional_solver_iterations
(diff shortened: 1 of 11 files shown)


sebcrozet:
@Ughuuu 2D contact manifold are already limited to up to 2 points, so no additional reduction is needed.

PR #899, opened 2025-11-13
Make Broadphase Serializable to Json

dog-molecule:
Serde can serialize almost the entire Rapier state to JSON; the only thing that's preventing full JSON-serialization (at least in my implementation) is this hashmap in Broadphase. Rapier already has serialization tools that can serialize hashmaps as vector tuples, so this PR just uses these existing utilities.


sebcrozet:
The docs CI failure is unrelated to this change and has been fixed on `main`.

PR #924, opened 2026-03-07
Perf: optimize kinematic body velocity interpolation by tracking them…

yagizgil:
This PR optimizes the kinematic body velocity interpolation by tracking active kinematic bodies within each island. Previously, the pipeline iterated over all active bodies to find kinematic ones. This change improves performance in scenes with many dynamic bodies and few kinematic ones. Closes the internal TODO in physics_pipeline.rs.


sebcrozet:
Thank you for this PR! The island system has changed a lot since then so this PR is no longer applicable. If a similar optimization is still needed please open a new issue or PR, and provide a benchmark motivating the change.

PR #929, opened 2026-04-15
fix(ccd): honor filter_contact_pair PhysicsHook during CCD (closes #754)

### LINKED ISSUE #754: CCD Filtering is not implemented
Right now, filtering in rapier is done in narrow phase only, for contacts and intersections (bodies and areas).
The problem is in case of CCD we don't do any of these filterings, only after it's too late, arriving at cases like these:
https://github.com/appsinacup/godot-rapier-physics/issues/103

The solution would be to either call the `filter_contact_pair` in the CCD solver:
- `ccd_solver.rs` function `find_first_impact`: This function only takes into account collision bits, so right now it will intersect but generate no contacts, and there is no way to fix this. Another option would be adding another function for custom filtering of CCD collisions.

nickyvanurk:
## What

Makes Continuous Collision Detection consult the user's `PhysicsHooks::filter_contact_pair` hook, matching narrow-phase semantics. Previously, CCD would clamp a fast-moving body's motion at a predicted impact with a pair the user had filtered out, leading to the bug described in #754.

## Why

Per the issue, *"filtering in rapier is done in narrow phase only, for contacts and intersections (bodies and areas). The problem is in case of CCD we don't do any of these filterings."* When a user enables CCD on an object and also installs a `filter_contact_pair` hook to suppress certain collisions, the hook is bypassed during CCD's time-of-impact sweeps — so the fast body still gets motion-clamped on a pair the user told the engine to ignore.

## How

- New private helper `pair_filtered_out_by_hooks` in `src/dynamics/ccd/ccd_solver.rs` that mirrors the filter call in `NarrowPhase::compute_contacts` (checks `ActiveHooks::FILTER_CONTACT_PAIRS` on either collider, builds a `PairFilterContext`, returns `true` on `None`).
- Called at all three pair-evaluation sites in the CCD solver:
  - `find_first_impact` (substep splitting)
  - `predict_impacts_at_next_positions` first-sweep loop (initial TOI collection)
  - `predict_impacts_at_next_positions` resweep loop (after freezing bodies)
- `hooks: &dyn PhysicsHooks` is threaded through the two public `CCDSolver` methods and through `PhysicsPipeline::run_ccd_motion_clamping`.

## Breaking change

`CCDSolver::find_first_impact` and `CCDSolver::predict_impacts_at_next_positions` now take an extra `hooks: &dyn PhysicsHooks` argument. `PhysicsPipeline::step` callers are **unaffected** — only code that drives the CCD solver directly (uncommon) needs to pass the hooks argument.

If you'd prefer a non-breaking additive approach (e.g., keep the current signatures and add `*_with_hooks` variants that take the extra param), I'm happy to restructure — I went with the breaking approach because rapier already makes breaking changes between minor versions and it keeps the surface clean.

## Test

Added `pipeline::physics_pipeline::test::ccd_respects_filter_contact_pair_hook` (3D) that:
1. Creates a 200 m/s CCD body and a stationary CCD body directly in its path
2. Installs a hook that rejects the pair
3. Steps 5 times
4. Asserts (a) the hook was called from CCD (since narrow phase never sees the pair at these speeds), and (b) the fast body flew through to `x > 10` rather than clamping near the stationary body (which happens without the fix at `x ≈ -0.77`).

Empirically verified on a downstream project that the same scenario fails before this patch and passes after.

## CHANGELOG

Added an `Unreleased` section with the fix and the breaking-change note.

## Notes for review

- **Sensor asymmetry**: the narrow phase invokes `filter_contact_pair` for sensor pairs too (and uses `solver_flags` to decide what to do downstream). CCD's sweep sites skip sensors *before* calling this helper, which is intentional — CCD only resolves contact TOIs, and sensor intersection events are reported elsewhere. Called out in the helper's doc comment.
- **Double-filter per step**: a filtered pair may get `filter_contact_pair` called up to 3× per step now (once per CCD sweep + narrow phase). Narrow phase already calls per-pair per-step; I don't expect the CCD sweeps to dominate in practice since CCD is only engaged for fast bodies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## Unreleased
+
+### Fixed
+
+- Continuous Collision Detection now consults the user's `PhysicsHooks::filter_contact_pair`
+  hook, matching narrow-phase semantics. Previously, CCD would clamp a fast-moving body's
+  motion at a predicted impact with a pair the user had filtered out. Closes #754.
+
+### Modified
+
+- **Breaking (CCD solver direct callers only):** `CCDSolver::find_first_impact` and
+  `CCDSolver::predict_impacts_at_next_positions` now take an extra `hooks: &dyn PhysicsHooks`
+  argument. `PhysicsPipeline::step` callers are unaffected.
+
 ## v0.31.0 (09 Jan. 2026)
 
 ### Modified
--- a/src/dynamics/ccd/ccd_solver.rs
+++ b/src/dynamics/ccd/ccd_solver.rs
@@ -1,13 +1,51 @@
 use super::TOIEntry;
 use crate::alloc_prelude::*;
 use crate::dynamics::{IntegrationParameters, IslandManager, RigidBodyHandle, RigidBodySet};
-use crate::geometry::{BroadPhaseBvh, ColliderParent, ColliderSet, CollisionEvent, NarrowPhase};
+use crate::geometry::{
+    BroadPhaseBvh, Collider, ColliderHandle, ColliderParent, ColliderSet, CollisionEvent,
+    NarrowPhase,
+};
 use crate::math::Real;
 use crate::parry::utils::SortedPair;
-use crate::pipeline::{EventHandler, QueryFilter};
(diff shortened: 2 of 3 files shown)


sebcrozet:
Looks good, thanks!

PR #939, opened 2026-06-02
fix aliasing UB, layout UB, and incorrect initialization UB

Morgane55440:
the `get_disjoint_mut`-likes are aliasing UB under the current memory model.
for more info, see the internal code of `get_disjoint_mut` and the relevant history.

`as_vector` on `RigidBodyVelocity` is UB because it is not `#[repr(C)]`

`reset_buffer` can cause incorrect initialization UB because it initializes `len` bytes, but it says it has initialized `len * size_of:::<T>` bytes.

the `get_disjoint_mut`-likes have been checked to result in the same assembly after fix : 
- https://godbolt.org/z/5TTfrs4zP
- https://godbolt.org/z/Tdbh75PT9

the asm for `reset_buffer` is slightly different, but it still uses `memset` which is the important part.

### DIFF
--- a/src/data/graph.rs
+++ b/src/data/graph.rs
@@ -146,12 +146,10 @@ fn index_twice<T>(arr: &mut [T], a: usize, b: usize) -> Pair<&mut T> {
     } else if a == b {
         Pair::One(&mut arr[max(a, b)])
     } else {
-        // safe because a, b are in bounds and distinct
-        unsafe {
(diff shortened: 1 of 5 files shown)


sebcrozet:
Thank you for this PR. Most (hopefully all) UB issues have been addressed in the latest version of rapier. We are now also running miri tests in CI.

Issue #925, opened 2026-03-09
vehicleController.wheelRotation(i) only updated when steering


tomo0613:
when calling `vehicleController.wheelRotation(i)` on a `RAPIER.DynamicRayCastVehicleController` it will return `0` until steering is applied, even if it is moving.

when pressing forward the wheels will NOT rotate. 
when pressing forward AND when turning to the right (pressing right) the front wheels will start rotate (in the correct direction), when turning left the wheels will rotate the OPPOSITE direction.

https://codepen.io/editor/Tomo0613/pen/019cd3e3-fe8b-738c-9211-e2529094714a
I added logging to the official ThreeJs demo (https://threejs.org/examples/?q=rapie#physics_rapier_vehicle_controller)
and reduced the wheel cylinder geometry's segment count to make it more visible.

PR #945, opened 2026-06-07
Fix issue 925 (wheel rotation calculation)

tomo0613:
https://github.com/dimforge/rapier/issues/925

explanation (by Claude):

The Bug
In `src/control/ray_cast_vehicle_controller.rs`, the wheel rotation update uses a `fwd` (forward) vector computed independently from the chassis rotation:
```
let mut fwd = chassis.position().rotation * Vector::ith(self.index_forward_axis, 1.0);
let proj = fwd.dot(wheel.raycast_info.contact_normal_ws);
fwd -= wheel.raycast_info.contact_normal_ws * proj;
```
But the friction impulse system in `update_friction()` computes forward as:
```
self.forward_ws[i] = surf_normal_ws.cross(self.axle[i]).normalize_or_zero();
```
These two methods produce vectors that can point in opposite directions depending on the axle axis convention. For example, with `index_forward_axis = 2` (Z forward) and a wheel axle along `-X`:

- `forward_ws[i]` = `Y × (-X)` = `-Z`
- `fwd` = `+Z` ← exact opposite!

The result: `delta_rotation` has the wrong sign, making the wheel appear to spin backward (and it cancels out to ~0). When steering is applied, the axle rotates, which partially corrupts/corrects the mismatch — which is exactly why rotation only appears when steering.


tomo0613:
video evidence of the bug:

https://github.com/user-attachments/assets/d7c0a710-ebf1-447c-879c-8eaf153db993

and the fix:

https://github.com/user-attachments/assets/ba77d963-525c-4235-a85c-aa6e1be326a3

tomo0613:
As I was testing, found out that the fix is not complete.
better than before ( video evidence :) ), but;
no rotation when wheels in the air and should spin
and wheels can NOT lock when braking

I'm comparing the ccde to cannon-es (also bullet port), and will update this PR

tomo0613:
I added 2 extra conditions based on cannon-es implementation.
https://github.com/pmndrs/cannon-es/blob/master/src/objects/RaycastVehicle.ts

this handles if a wheel should start spinning (if wheel in air and engine force is applied, or on ground but looses grip while engine force is applied [burnout]):
```
      if ((wheel.sliding || !wheel.isInContact) && wheel.engineForce !== 0 && wheel.useCustomSlidingRotationalSpeed) {
        // Apply custom rotation when accelerating and sliding
        wheel.deltaRotation = (wheel.engineForce > 0 ? 1 : -1) * wheel.customSlidingRotationalSpeed * timeStep
      }
```

this handles if a wheel should stop spinning:
```
      // Lock wheels
      if (Math.abs(wheel.brake) > Math.abs(wheel.engineForce)) {
        wheel.deltaRotation = 0
      }
```


I also checked ammo.js and these issues are there as well
https://github.com/kripken/ammo.js/

Issue #946, opened 2026-06-07
RayCastVehicle has infinite effective braking friction if it has 0 rotation values


tomo0613:
When a `RAPIER.DynamicRayCastVehicleController`'s rotation quaternion has 0 values, and applying brake force to wheels, (like using handbrake) the wheels will act as if they have much higher friction then normally.
The issue originates from Bullet physics and can be observed in the official cannon-es and three.js (rapier) examples.
Like when accelerating in a straight line (without any steering) to high enough speed, and using the handbrake, the wheels will act as if they were glued to the ground, and the vehicle will roll/tilt forward (depending on brake force).
However doing the same with just tapping left or right (steering in any direction for a brief moment), and then using the handbrake, the wheels will slip/slide as expected.

PR #947, opened 2026-06-07
fix wheel impulse scaling

tomo0613:
https://github.com/dimforge/rapier/issues/946

explanation by Claude:

The Bug: A misplaced gate in updateFriction

In `update_friction()`, after determining that `sliding == true` (the combined forward+side impulse exceeds the friction circle), there's a correction loop that scales both impulses down by `skid_info`. ...

```
// before
if wheel.side_impulse != 0.0 && wheel.skid_info < 1.0 {

// after
if wheel.skid_info < 1.0 {
```
The `wheel.side_impulse != 0.0` guard is the whole problem. It was presumably written as a micro-optimisation ("skip wheels with no lateral load"), but it incorrectly prevents `forward_impulse` from being scaled down when braking straight — since `side_impulse` is exactly `0.0` in that case, the entire body of the if is skipped, and the braking impulse is applied without any friction-circle capping.
The `skid_info < 1.0 `check alone is the correct and sufficient guard — it's only `< 1.0` when the friction circle is actually exceeded, so there's no wasted work. Multiplying `side_impulse` (which is already 0.0) by `skid_info` is a no-op, so removing the outer guard has zero downside.

This bug exists in the original Bullet source (lines 675–682 of `btRaycastVehicle.cpp`) and was presumably never caught because most demos and tutorials involve some steering before heavy braking — the exact scenario that masks it.

### DIFF
--- a/src/control/ray_cast_vehicle_controller.rs
+++ b/src/control/ray_cast_vehicle_controller.rs
@@ -659,7 +659,7 @@ impl DynamicRayCastVehicleController {
 
         if sliding {
             for wheel in &mut self.wheels {
-                if wheel.side_impulse != 0.0 && wheel.skid_info < 1.0 {
+                if wheel.skid_info < 1.0 {
                     wheel.forward_impulse *= wheel.skid_info;
                     wheel.side_impulse *= wheel.skid_info;
                 }


sebcrozet:
Thank you for this PR. Part of this was integrated to #980
Feel free to open an issue if the problem persists.

PR #950, opened 2026-06-21
Fix issue 925 v2

### LINKED ISSUE #925: vehicleController.wheelRotation(i) only updated when steering

tomo0613:
(messed up my previous PR https://github.com/dimforge/rapier/pull/945 )
These changes should fix https://github.com/dimforge/rapier/issues/925
And handle 2 more scenarios, to make the wheels react to brake and engine forces.

Changes are based on cannon-es lib.

### DIFF
--- a/python/tests/test_controllers.py
+++ b/python/tests/test_controllers.py
@@ -417,3 +417,107 @@ def test_wheel_tuning_default_kwargs(ns):
     t2 = ns.WheelTuning(suspension_stiffness=10.0, friction_slip=5.0)
     assert abs(t2.suspension_stiffness - 10.0) < 1e-5
     assert abs(t2.friction_slip - 5.0) < 1e-5
+
+def test_wheel_rotation_updates_without_steering(ns):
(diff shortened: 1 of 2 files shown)


sebcrozet:
Thank you for this PR. Part of this was integrated to #980
Feel free to open an issue if the problem persists.

PR #962, opened 2026-07-13
Ensure `sh` prefix is present in the package.json scripts where needed

PoseidonEnergy:
I sometimes use a Cygwin terminal on Windows to build Rapier. When I run `npm run build` in the `typescript/rapier-compat` directory, it fails on `./build-rust.sh` because it doesn't have the `sh` prefix. This PR adds the `sh` prefix where needed.


sebcrozet:
Thanks!

PR #963, opened 2026-07-14
Enforce LF line endings for shell scripts

PoseidonEnergy:
This PR adds a `.gitattributes` rule to enforce Unix-style LF line endings for all shell scripts:

```gitattributes
*.sh text eol=lf
```
I sometimes use a Cygwin terminal on Windows to build Rapier. When I run `./builds/prepare_builds/prepare_all_projects.sh` or `./builds/prepare_builds/build_all_projects.sh` from within `./typescript` it fails because of the unexpected line endings.

No functional code changes. This only standardizes line endings for shell scripts.


sebcrozet:
Thank you!

PR #965, opened 2026-07-14
Clean Cargo artifacts between WASM variants

PoseidonEnergy:
This PR adds a `cargo clean` step before each `wasm-pack` build in `rapier-compat`.

The build variants share the same Cargo target directory but use different `RUSTFLAGS`, particularly for SIMD and non-SIMD builds. Without cleaning, for me, Cargo reuses artifacts compiled with incompatible target features, causing non-SIMD builds to contain SIMD instructions and fail during `wasm-opt` validation.

Cleaning before each variant ensures every WASM package is rebuilt with the correct compiler flags.

### DIFF
--- a/typescript/rapier-compat/build-rust.sh
+++ b/typescript/rapier-compat/build-rust.sh
@@ -47,4 +47,13 @@ else
     export additional_rustflags=''
 fi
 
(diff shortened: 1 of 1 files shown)


sebcrozet:
Thanks!

PR #971, opened 2026-08-02
Specialize twist friction accumulation for small contact counts

fyrsta7:
## Summary

This specializes twist-friction tangent impulse accumulation for constraints with one or two normal contact parts, while keeping the existing loop as the fallback for other counts.

## Why this can improve performance

Contact constraints commonly have a very small number of normal parts. Handling the one- and two-part cases directly reduces loop overhead and repeated indexing on that solver path without changing the arithmetic performed for each tangent impulse.

## Validation

- `git diff --check`
- `cargo check -p rapier3d`

## Benchmark

Current machine, release microbenchmark, 7 samples, median:

| Benchmark | Base | This change | Delta |
|---|---:|---:|---:|
| 30 physics pipeline steps with contact constraints | 638,331.425 ns | 632,754.625 ns | 0.87% faster |


fyrsta7:
Closing this draft because follow-up benchmarking did not show a large enough performance gain to justify this as an upstream performance PR.

PR #973, opened 2026-08-02
feat: significant performance and stability improvements

sebcrozet:
This PR introduces significant performance an stability improvements. See individual commit messages and the changelog for details. In particular:
- This implements intra-island parallelism. The `parallel` feature is now worth enabling for all types of scenes, even smaller ones where the overhead was prohibitive in previous versions of Rapier. (Note that I used Claude Code to help with testing and fixing soundness issues & miri tests).
- Removed the `simd-stable` and `simd-nightly` features. Simd is now always enabled (and automatically fallbacks to scalar on unsupported targets).
- Cross-platform determinism is now compatible with both SIMD and parallelism. A simulation on single-thread WASM remains bit-identical to the same multi-thread SIMD-accelerated native simulation, assuming the `enhanced-determinism` feature is enabled.
- Enabling `enhanced-determinism` no longer has any measurable overhead (at least not on my Macbook Pro M4 Max).
- Don’t panic when a NaN happens in the simulation. Instead the problematic bodies are automatically disabled.
- Rework the CCD logic to something similar to Box2D v3’s approach. The rationale is that their CCD implementations work very well and with very low overhead. So we are starting back from  that version and will improve on top of it in the future.
- General stability improvements (part of it thanks to matching some of the Box3D default for the constraint spring coefficients).
- Added IntelliJ/RustRover runfiles.

The graphs below show performance comparisons of various 2D and 3D physics engines in single-thread and multi-thread mode. The benchmarks are run on a Macbook Pro M4 Max. The general trends are that the new rapier is:
- about 50% (single-thread) or 300% (multi-thread) faster than the previous version.
- about on-par-to-slightly-faster than Box3D (especially on single-threaded simulations).
- about on-par-to-slightly-slower than Box2D (especially on multi-threaded simulations).
- about 150% (single-thread) or 280% (multi-thread) faster than Avian.
- about 50% (single-thread) or 100% (multi-thread) faster than Jolt. It is also significantly more stable on stacks than Jolt.

**NOTE:** the benchmark code was entirely generated by AI (with manual validation of some of them). In addition, the Avian solver step count was set to 4 instead of the default 6, to match the solver count of Box2D/3D and Rapier. All parallel benchmarks use 8 threads.

3d_balls
3d_boxes
3d_capsules
3d_convex_hulls
3d_joint_grid
3d_large_pyramid
3d_many_pyramids
3d_ragdolls

2d_balls
2d_boxes
2d_capsules
2d_convex_polygons
2d_joint_grid
2d_large_pyramid
2d_many_pyramids
2d_ragdolls


sebcrozet:
The python bindings will be fixed in a follow-up PR.

PR #988, opened 2026-08-14
Add criterion benchmarks for steady-state pipeline stepping (all four crates)

marknefedov:
Adds a `harness = false` criterion bench with two scenes stepped in steady state: a 210-box pyramid (persistent convex contacts, contact-constraint update path) and 300 mixed convex bodies on trimesh terrain (contact clustering, trimesh manifold path). Bodies have sleeping disabled and each scene settles for 150 to 200 steps outside the timed region, so every timed step does full solver work.

The 3d bench body is written against `Real` and reused verbatim by rapier3d-f64 through a two-line extern-crate alias shim. rapier2d has a 2D-native body (2D pyramid, polyline terrain), shared the same way with rapier2d-f64.

Placement is deterministic (small LCG, fixed seed). No new dependencies beyond criterion as a dev-dependency. Runs on stable. On an M-series MacBook the f32 builds settle around 260 us per step for the pyramid and 460 us for the terrain scene; the f64 builds are roughly 1.7x that.

### DIFF
--- a/crates/rapier2d-f64/Cargo.toml
+++ b/crates/rapier2d-f64/Cargo.toml
@@ -112,9 +112,15 @@ nalgebra = { workspace = true }
 glamx = { workspace = true, features = ["nalgebra"] }
 
 [dev-dependencies]
+criterion = "0.8"
 bincode.workspace = true
 serde_json.workspace = true
 serde = { workspace = true, features = ["std"] }
 oorandom.workspace = true
 
 [target.'cfg(target_vendor = "apple")'.dependencies]
+
+[[bench]]
+name = "pipeline"
+path = "benches/pipeline.rs"
+harness = false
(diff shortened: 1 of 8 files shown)


Ughuuu:
This adds a performance test that needs to be manually run or is automatic? I am not familiar with criterion but what is it trying to solve? (It's am not a maintainer of rapier, just curious, sorry if unrelated).

marknefedov:
I'm mostly using it to catch regressions while tinkering or optimize stuff, it reruns benchmarks a ton of times, to make them statistically meaningful and outputs a diff to previous runs.
It can be added to CI to do automated testing.
here is an example: https://criterion-rs.github.io/book/user_guide/comparing_functions.html

Ughuuu:
Interesting, and in this PR what is it checking for? It checks that there isn't a performance regression? Or it compares functions to another for speed?

marknefedov:
Here I just add harness configuration so it is available locally, it can be added to CI, but AFAIK dim forge uses its own bot.
I used it to check validate that [some optimizations](https://github.com/dimforge/parry/pull/441) were working as excepted and worth create a PR for.

PR #989, opened 2026-08-14
Per-step allocation and inner-loop cleanups in the step pipeline

marknefedov:
Four changes removing per-step allocator traffic and one loop-invariant lookup:

- `carry_warmstart_data`: the manifold-level `subshape_pos1()` lookup moved out of the per-point matching loop. It is invariant there and always `None` for cluster manifolds, but was re-checked for every candidate point of an O(prev_points x target_points) loop.
- Narrow phase: the serial pair-transition buffer is a persistent `NarrowPhase` scratch field (take and restore), so its capacity survives across steps. Same idiom the function already uses for `update_candidates` and `solver_graph_dirty`.
- Substep: the joint island event `drain().collect()` replaced with the take-and-restore idiom used three lines below for multibody chain events. Removes a per-substep Vec.
- Staged island solver: the two generic multibody solver velocity vectors are resized in place and zeroed instead of rebuilt with `DVector::zeros` every step. Reallocation only happens when the total DoF count grows.

On the benchmark scenes from #988 the effect was within noise (-1.6% on the pyramid at p = 0.05, wider on the terrain scene). Withdrawn for lack of demonstrable numbers.

### DIFF
--- a/src/dynamics/solver/staged_island_solver/helpers.rs
+++ b/src/dynamics/solver/staged_island_solver/helpers.rs
@@ -12,7 +12,6 @@ use crate::dynamics::solver::solver_contact_graph::SolverContactGraph;
 #[cfg(feature = "dim3")]
 use crate::dynamics::solver::velocity_solver::GyroParams;
 use crate::dynamics::{IntegrationParameters, MultibodyJointSet, RigidBodyHandle, RigidBodySet};
-use crate::math::DVector;
 use parry::math::SIMD_WIDTH;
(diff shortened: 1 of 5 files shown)


marknefedov:
Closing this one: the measured effect on the benchmark scenes was within noise, and I'd rather not ask for review time on changes I can't demonstrate. The benchmarks in #988 stand on their own; if a scene shows these allocations mattering I'll reopen with numbers.

PR #994, opened 2026-08-16
Fix WASM leak when restoring snapshots

### LINKED ISSUE #990: TypeScript: World.restoreSnapshot() leaks ~1730 bytes per call regardless of world size
`World.restoreSnapshot()` leaks a fixed ~1730 bytes of WASM memory per call, even when the returned world is `free()`d.

The leak is **independent of world size**, which suggests a fixed-size allocation per restored `World` rather than anything proportional to the snapshot contents.

(Filed here rather than `rapier.js`, which is archived and points to `typescript/`.)

### Environment

- `@dimforge/rapier2d-deterministic-compat` 0.20.0
- Node v25.4.0, macOS 15 (darwin/arm64)

### Reproduction

```js
// node --expose-gc leak-repro.mjs [cycles] [bodies]
import RAPIER from "@dimforge/rapier2d-deterministic-compat";

const CYCLES = Number(process.argv[2] ?? 4000);
const BODIES = Number(process.argv[3] ?? 40);
await RAPIER.init();

function buildWorld() {
  const world = new RAPIER.World(new RAPIER.Vector2(0, -9.81));
  const ground = world.createRigidBody(RAPIER.RigidBodyDesc.fixed().setTranslation(0, -1));
  world.createCollider(RAPIER.ColliderDesc.cuboid(50, 1), ground);
  for (let i = 0; i < BODIES; i++) {
    const b = world.createRigidBody(
      RAPIER.RigidBodyDesc.dynamic().setTranslation((i % 8) * 1.1, 1 + Math.floor(i / 8) * 1.1),
    );
    world.createCollider(RAPIER.ColliderDesc.cuboid(0.5, 0.5).setDensity(1), b);
  }
  for (let i = 0; i < 60; i++) world.step();
  return world;
}

const sample = () => (global.gc(), global.gc(), process.memoryUsage());

const source = buildWorld();
const snapshot = source.takeSnapshot();

// Warm up so one-time allocations are not counted as growth.
for (let i = 0; i < 200; i++) RAPIER.World.restoreSnapshot(snapshot).free();

const before = sample();
for (let i = 0; i < CYCLES; i++) {
  const w = RAPIER.World.restoreSnapshot(snapshot);
  w.step();
  w.free();
}
const after = sample();
console.log(`external: ${((after.external - before.external) / CYCLES).toFixed(1)} B/cycle`);

// Control: step + takeSnapshot on one world, no restore.
const control = buildWorld();
const cBefore = sample();
for (let i = 0; i < CYCLES; i++) { control.step(); control.takeSnapshot(); }
const cAfter = sample();
console.log(`control:  ${((cAfter.external - cBefore.external) / CYCLES).toFixed(1)} B/cycle`);
```

### Results

`external` is where WASM linear memory shows up in Node's accounting. 2500 cycles each:

| bodies | snapshot size | `restoreSnapshot` + `free` | control (`step` + `takeSnapshot`) |
|---:|---:|---:|---:|
| 5 | 7,947 B | **1730.2 B/cycle** | 0.0 B/cycle |
| 40 | 60,114 B | **1730.2 B/cycle** | 26.2 B/cycle |
| 300 | 441,562 B | **1730.2 B/cycle** | 180.0 B/cycle |

The snapshot size spans 55× while the leaked amount is byte-identical, and the control path — which exercises stepping and serialization but never restores — is essentially flat. Over 4000 cycles at 40 bodies the process also grows ~48 MB RSS and ~240 B/cycle of JS heap.

### Why this matters

This shows up in rollback netcode, where `restoreSnapshot` is called on every correction. At a realistic 1–2 rollbacks/second that is roughly 6–13 MB/hour of unbounded growth — slow, but it never plateaus, so a long session eventually dies. There is no workaround from the JS side: the returned world is freed, and nothing is retained by the caller.

Happy to test a patch or provide more instrumentation if useful.

Bardakor:
Fixes #990.

`World.fromRaw()` transferred all nine physics components out of `RawDeserializedWorld` but left the drained wasm-bindgen wrapper allocated. That container accounts for the measured fixed growth per restore. `World.restoreSnapshot()` also created a temporary `SerializationPipeline` without honoring its documented manual-free contract, although independent ablation found no measurable growth from that temporary.

This change releases both temporary owners in `finally` blocks after deserialization. The transferred components remain owned by the returned `World`, and cleanup also runs if construction or deserialization throws.

The regression test captures the module's actual `WebAssembly.Memory` and disables finalizer callbacks before initialization. After 200 warm-up restores, it restores and explicitly frees 2,500 empty worlds. On the unpatched code, the 2D buffer grew from 1,572,864 to 5,898,240 bytes and the 3D buffer grew to 6,291,456 bytes. With both cleanup paths, neither buffer grows.

## Validation

- `npm run build` in `typescript/rapier-compat` for all six Rust/WASM and JavaScript compatibility builds
- `npm test -- --runInBand` in `typescript/rapier-compat`: 5 suites and 8 tests passed
- `npx prettier --check src.ts/pipeline/world.ts rapier-compat/tests/WorldMemory.test.ts`
- `cargo fmt -- --check` in `typescript`

### DIFF
--- a/typescript/rapier-compat/tests/WorldMemory.test.ts
+++ b/typescript/rapier-compat/tests/WorldMemory.test.ts
@@ -1,0 +1,93 @@
+interface RestorableWorld {
+    free(): void;
+}
+
+interface SnapshotWorld extends RestorableWorld {
+    takeSnapshot(): Uint8Array;
+}
+
+interface RapierApi {
+    init(): Promise<void> | void;
+}
+
+class NoopFinalizationRegistry {
+    constructor(_callback: unknown) {}
+
+    register(_target: object, _heldValue: unknown, _token?: object) {}
+
+    unregister(_token: object) {
+        return true;
+    }
+}
+
+async function initializeWithCapturedMemory<T extends RapierApi>(
+    load: () => Promise<T>,
+): Promise<{rapier: T; memory: WebAssembly.Memory}> {
+    const originalFinalizationRegistry = (globalThis as any)
+        .FinalizationRegistry;
+    const originalInstantiate = WebAssembly.instantiate;
+    let memory: WebAssembly.Memory | undefined;
+
+    (globalThis as any).FinalizationRegistry = NoopFinalizationRegistry;
+    (WebAssembly as any).instantiate = async (...args: any[]) => {
+        const result = await (originalInstantiate as any)(...args);
+        const instance =
+            result instanceof WebAssembly.Instance ? result : result.instance;
+        if (instance.exports.memory instanceof WebAssembly.Memory) {
+            memory = instance.exports.memory;
+        }
+        return result;
+    };
+
+    try {
+        const rapier = await load();
+        await rapier.init();
+        if (!memory) throw new Error("Rapier WASM memory was not captured");
+        return {rapier, memory};
+    } finally {
+        (globalThis as any).FinalizationRegistry = originalFinalizationRegistry;
+        WebAssembly.instantiate = originalInstantiate;
+    }
+}
+
+function expectRestoresToReuseWasmMemory(
+    source: SnapshotWorld,
+    memory: WebAssembly.Memory,
+    restore: (snapshot: Uint8Array) => RestorableWorld,
+) {
+    const snapshot = source.takeSnapshot();
+    source.free();
+
+    for (let i = 0; i < 200; i++) restore(snapshot).free();
+    const bytesBefore = memory.buffer.byteLength;
+
+    for (let i = 0; i < 2500; i++) restore(snapshot).free();
+
+    expect(memory.buffer.byteLength).toBe(bytesBefore);
+}
+
+test("2d snapshot restores reuse WASM memory after worlds are freed", async () => {
+    const {rapier, memory} = await initializeWithCapturedMemory(
+        () => import("../builds/2d-deterministic/pkg"),
+    );
+    const source = new rapier.World(new rapier.Vector2(0, -9.81));
+
+    expectRestoresToReuseWasmMemory(
+        source,
+        memory,
+        rapier.World.restoreSnapshot,
+    );
+});
+
+test("3d snapshot restores reuse WASM memory after worlds are freed", async () => {
+    const {rapier, memory} = await initializeWithCapturedMemory(
+        () => import("../builds/3d-deterministic/pkg"),
+    );
+    const source = new rapier.World(new rapier.Vector3(0, -9.81, 0));
+
+    expectRestoresToReuseWasmMemory(
+        source,
+        memory,
+        rapier.World.restoreSnapshot,
+    );
+});
--- a/typescript/src.ts/pipeline/world.ts
+++ b/typescript/src.ts/pipeline/world.ts
@@ -188,17 +188,21 @@ export class World {
     public static fromRaw(raw: RawDeserializedWorld): World {
         if (!raw) return null;
 
-        return new World(
-            VectorOps.fromRaw(raw.takeGravity()),
-            raw.takeIntegrationParameters(),
-            raw.takeIslandManager(),
-            raw.takeBroadPhase(),
-            raw.takeNarrowPhase(),
-            raw.takeBodies(),
-            raw.takeColliders(),
-            raw.takeImpulseJoints(),
-            raw.takeMultibodyJoints(),
-        );
+        try {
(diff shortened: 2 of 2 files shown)


jounih:
Tested — but first, **I need to correct my previous comment.** The 26.2 B/cycle residual I reported for `fromRaw`-only is not real, and since you cited it as the justification for including the `SerializationPipeline` cleanup, the correction matters.

26.2 × 2500 cycles = 65,500 B — one 64 KiB WASM page. It was a single page of allocator noise at one measurement point, not a rate. Varying the cycle count shows it immediately (300 bodies, `fromRaw` free only):

| cycles | total `external` delta | "per cycle" |
|---:|---:|---:|
| 2,500 | +65,576 B | 26.2 |
| 5,000 | −376,018 B | −75.2 |
| 10,000 | −310,482 B | −31.0 |
| 20,000 | −113,874 B | −5.7 |

It goes *negative*. My mistake was repeating the same 2,500-cycle run and treating "reproducible to the decimal" as evidence of a rate — identical runs cross the same page boundary every time, so of course the number was stable. That is reproducibility, not linearity.

The control shows the method can tell the difference — unpatched, the real leak holds its rate as the run grows:

| cycles | unpatched, 300 bodies |
|---:|---:|
| 2,500 | 1730.2 B/cycle |
| 10,000 | 1729.3 B/cycle |
| 20,000 | 1707.6 B/cycle |

**So: `raw.free()` in `fromRaw` is the entire leak.** The `SerializationPipeline` free still belongs in this PR — the class documents a manual-free contract and the temporary plainly violates it — but as lifecycle correctness, not because it recovers measurable memory. I should not have claimed otherwise.

### New data: 3D, which I had not measured before

Both compat builds, released `0.20.0` artifacts, patched in place with exactly this PR's two `finally` blocks:

| build | unpatched | both patches |
|---|---:|---:|
| 2d-deterministic | 1730.2 B/cycle | **0.0** |
| 3d-deterministic | 1861.2 B/cycle | **0.0** |

Flat at 5, 40 and 300 bodies, and still flat at 20,000 cycles. 3D leaks a different fixed constant (1861.2 vs 1730.2, and 1887.5 at 5 bodies), consistent with a larger `RawDeserializedWorld` — so your decision to cover both builds in the regression test is warranted, and the 3D figure was previously unmeasured by anyone.

### Determinism suite

265 tests against both patched builds: 263 passed. The two failures are ours, not yours — a long soak and a `SharedArrayBuffer` concurrency test, both timing-sensitive, both passing in isolation, and one of them also fails intermittently on the *pristine* baseline. Neither touches the snapshot path.

What matters for this PR passed in both configurations: golden state checksums pinned on arm64 and compared bit-for-bit, snapshot/restore round-trips, and post-restore resimulation. The change is behaviour-preserving.

### On your test

Disabling `FinalizationRegistry` with a no-op is the right call and worth keeping — it forces the test to measure *explicit* ownership rather than whatever the GC happens to reclaim, which is exactly the contract being fixed. Asserting on `memory.buffer.byteLength` is also more robust than the `process.memoryUsage().external` I was using: page-granular, but immune to precisely the noise that misled me above.

One caveat on my side: I patched the published `dist/rapier.mjs` rather than building your branch, so this validates the *change*, not the build. The transformation is byte-for-byte the minified equivalent of your two `finally` blocks.

Bardakor:
Thanks for correcting this with the cycle-count sweep. You're right: the repeatable 65,576 B movement was one page-boundary effect, not evidence of a 26.2 B/cycle residual. I relied on that figure in my earlier reply, and that attribution was wrong. `raw.free()` accounts for the entire measured linear growth.

I've corrected the PR description to separate the two changes. The `SerializationPipeline` `finally` remains because `restoreSnapshot()` creates an owned wrapper whose API explicitly requires manual cleanup. That is lifecycle correctness, not measurable memory recovery.

The new 3D results support keeping both compat builds in the regression test: they show a distinct unpatched slope and zero growth with the patch across body counts and through 20,000 cycles. The determinism run also covers the behavioral risk here. Golden checksums, snapshot round-trips, and post-restore resimulation passed in both configurations. Given that the two timing-sensitive failures pass in isolation and one also occurs on the pristine baseline, I don't consider them regressions from this change.

The artifact caveat is noted too: this independently validates the equivalent transformation against the released 0.20.0 bundles, while the branch build and CI validate the submitted source.

PR #999, opened 2026-08-23
Add `NarrowPhase::set_query_dispatcher`, skip serde for contact workspaces

Jondolf:
Adds `NarrowPhase::set_query_dispatcher`. This is useful for snapshot restoration, because (de)serialization drops custom query dispatchers, and there's no other pre-existing way to set it without constructing a new `NarrowPhase` (as far as I know). 

Also changes contact pairs to skip serialization for the contact workspace, which is transient and makes deserialization fail with `"Cannot deserialize custom shape"` for snapshots that contain custom shape contacts.


sebcrozet (approved):
Thank you!

sebcrozet (changes requested):
Actually, we shouldn’t skip the serialization of `workspace` as that could break determinism if the workspace contains warmstarting for some internal algorithm.

Instead, the deserialization panic should be addressed (I’m thinking we should have a mechanism that sets the workspace to a dummy value for unknown types. The specific contact algorithm would then automatically overwrite that dummy value with its own proper workspace type at its next execution.)

Jondolf:
Good point, I've reverted that part and scoped this PR to only add `set_query_dispatcher`. I opened https://github.com/dimforge/parry/pull/450 to address the deserialization problem separately, with a dummy workspace as you suggested.

PR #1000, opened 2026-08-27
This is a special edition; if the community needs it, they can add specific features themselves.

baka4n:
Just add a signature for me.


sebcrozet:
Hi! Thank you for this PR. However I don’t quite understand the point of this pull-request. Is this just a fork you are working with?
In any case, I’m closing it as this doesn’t seem to be something looking to be upstreamed.
Please let me know if I misunderstood the intent.

PR #1004, opened 2026-09-04
Fix character controller manifold collection

KaiStevenson:
Fixes an issue in which character controller collisions with multiple colliders, at least one of which is compound, cause a panic.

Parry query dispatcher `contact_manifolds` can `mem::take` or otherwise reorder or remove colliders from the `manifolds` arg. The code in the character controller assumed it was append only--in these cases it could cause incorrect computation, or in cases with a composite collider, a panic.

Fix by building a new array for each call.

Fable 5.1 was used for regression test only


sebcrozet (approved):
Looks good, thank you!

PR #1007, opened 2026-09-08
Make `CoefficientCombineRule::combine` public

Jondolf:
Combining coefficients can be useful in external applications too, so it's good to expose an official API for it.


sebcrozet (approved):
Thank you!

